PR #7, opened 2015-07-24
* Adds coveo-input-slider component to Styleguide

flehoux:


config/dependencies.js
```diff
@@ -1,4 +1,5 @@
 exports.dependencies = [
     'lib/js/collapse.js',
+    'lib/js/colorSlider.js',
```
wfortin: I'd call that `slider.js`

lib/js/colorSlider.js
```diff
@@ -0,0 +1,61 @@
...
+            'input[type=range]' + slider + '::-ms-fill-upper {',
+            'background: ',
+            colors.upper,
+            '}'
+        ].join('');
+
+        style.html([webkit, firefox, ie].join(''));
+    }
+
+    $.fn.colorSlider = function () {
```
wfortin: same here `slider`

wfortin:
Other than that 🚢it

PR #8, opened 2015-10-19
Add css comb

wfortin:
CSS Comb settings following codeguide.co rules.
Changes include settings for CSSComb, a gulp task to run comb, and modified files with new code style.

I'll add a watcher to run this on save.

I suggest checking the review with "split" view.


JPLachance:
Ship it!

less/controls/checkboxes.less
```diff
@@ -40,107 +40,111 @@
...
+
+        box-sizing: content-box;
+        display: block;
+        // Make it a small rectangle so the border will create an L-shape
+        width: 4px;
+        height: 8px;
+
+        content: '';
+        // Add a white border on the bottom and left, creating that 'L'
+
```
Unetitebiere: Enlever l'espace

wfortin: Not sure i understand...

package.json
```diff
@@ -30,6 +30,7 @@
     "gulp-autoprefixer": "3.0.2",
     "gulp-cheerio": "0.6.2",
     "gulp-concat": "2.6.0",
+    "gulp-csscomb": "^3.0.6",
```
marilynechan: On veut vraiment ^ ?

less/controls/checkboxes.less
marilynechan: Comments look a bit mixed up because of the properties reordering.

wfortin: Oh ok i get it, it seems that CSSComb does not keep the comments order, making it unrelated to the property. I'll fix that.

pastjean:
Would be nice if the build failed when the files checked-in aren't combed

1. Add a task testComb to gulp
2. Compare files
3. fail if they are different

marilynechan:
We already talked about it, but it would also be nice if the css was *combed* on save.
Otherwise, Ship It!

Unetitebiere:
LGTM! Ship it

pastjean:
+1 @marilynechan watch comb task + build fail task if someone forgets the watch save , otherwise :shipit:

PR #9, opened 2015-10-19
Switch all LESS files to SCSS

wfortin:
Thanks to [less2sass](https://www.npmjs.com/package/less2sass)
Had some cleaning up to do a little after, but everything seems fine !


scss/common/common.scss
```diff
@@ -1,4 +1,4 @@
-@import (reference) 'font';
```
marilynechan: This was detected as a file rename/move, but not colors.less to colors.scss. Please do a hg rename for history purpose!

pastjean: hg :trollface:

marilynechan:
All looks good to me, except that a few files do not seem to have been renamed.

JPLachance:
Look's good to me!

PR #11, opened 2015-10-27
* New styleguide colors adapted to branding

wfortin:


marilynechan:
Looks good to me. As @GermainBergeron suggested, it might be a good idea to give samples of the text-x and fill-x classes.

wfortin:
I've added #12 to make sure we don't forget to document it once we do the icon utility classes.

PR #13, opened 2015-10-28
Login splash screen styles with alignment and flex utility classes

wfortin:
Added basscss's utility-layout classes and picked a couple of flex utility classes also.

Added h1 and .h1 equivalent classes for font sizing.

Splash screen classes.

Demo here : http://codepen.io/wfortin/full/meLLLL/
Next step is implementing it in the authorization server.


scss/elements/buttons.scss
```diff
@@ -64,6 +64,8 @@
 }
 
 .btn {
+  position: relative;
```
JPLachance: Why setting position relative in the btn class and not in the btn-login class instead?

JPLachance:
Ship it!

resources/images/polyart.svg
```diff
@@ -0,0 +1 @@
...33L-304,688L-16,617Z" fill="#c1c1c1" stroke="#c1c1c1" stroke-width="1.51"/><path d="M-156,-405L-323,-346L-68,-192Z" fill="#f4f4f4" stroke="#f4f4f4" stroke-width="1.51"/><path d="M-141,1249L-333,1350L-65,1374Z" fill="#838383" stroke="#838383" stroke-width="1.51"/><path d="M-323,-346L-290,-63L-68,-192Z" fill="#f2f2f2" stroke="#f2f2f2" stroke-width="1.51"/><path d="M-150,982L-406,1201L-141,1249Z" fill="#969696" stroke="#969696" stroke-width="1.51"/><path d="M-290,-63L-229,246L-109,86Z" fill="#e3e3e3" stroke="#e3e3e3" stroke-width="1.51"/><path d="M-257,898L-406,1201L-150,982Z" fill="#a3a3a3" stroke="#a3a3a3" stroke-width="1.51"/><path d="M-406,1201L-333,1350L-141,1249Z" fill="#8b8b8b" stroke="#8b8b8b" stroke-width="1.51"/><path d="M-290,-63L-282,433L-229,246Z" fill="#dedede" stroke="#dedede" stroke-width="1.51"/><path d="M-304,688L-406,1201L-257,898Z" fill="#aaaaaa" stroke="#aaaaaa" stroke-width="1.51"/><path d="M-323,-346L-406,1201L-290,-63Z" fill="#dddddd" stroke="#dddddd" stroke-width="1.51"/><path d="M-290,-63L-406,1201L-282,433Z" fill="#c8c8c8" stroke="#c8c8c8" stroke-width="1.51"/><path d="M-406,1201L-304,688L-282,433Z" fill="#b6b6b6" stroke="#b6b6b6" stroke-width="1.51"/></svg>
```
marilynechan: Can you format it please?

scss/controls/splash.scss
```diff
@@ -0,0 +1,53 @@
...
+    width: 200px;
+    height: auto;
+    margin: 15px auto 60px;
+  }
+}
+
+.splash-box {
+  text-align: center;
+
+  background: $white;
```
marilynechan: What's our standard regarding this? We have utility classes `bg-white`, when should we use them and when should we set a background directly?

scss/controls/splash.scss
```diff
@@ -0,0 +1,53 @@
...
+
+  .title {
```
marilynechan: I'm giving a try at naming our classes like they do here: https://gist.github.com/bobbygrace/9e961e8982f42eb91b80#2-components .
.splash-box-title
.splash-box-description

scss/elements/color-dot.scss
```diff
@@ -0,0 +1,20 @@
+.color-dot {
+  display: inline-block;
+
+  border-radius: 100%;
+}
+.color-dot.critical, .color-dot.stopped {
```
marilynechan: Again with the Trello CSS Guide - `.color-dot.state-critical`?

scss/elements/color-dot.scss
```diff
@@ -0,0 +1,20 @@
...
+  background-color: $red;
+}
+.color-dot.major {
+  background-color: $orange;
+}
+.color-dot.minor {
+  background-color: $yellow;
+}
+.color-dot.none, .color-dot.normal {
```
marilynechan: Should this be the default bg color of `.color-dot`?
Also, I'm not sure we want many states with the same color. Shouldn't we map our internal states to a single `color-dot` state?

scss/elements/icons.scss
```diff
@@ -0,0 +1,30 @@
...
+
+  svg {
+    width: 100%;
+    height: 100%;
+  }
+}
+
+.icon-small {
+  width: 17px;
+  height: 17px;
```
marilynechan: Why not 18?

scss/elements/login-buttons.scss
```diff
@@ -0,0 +1,41 @@
+$btn-login-height: 54px;
```
marilynechan: I'm not sure this should go in the 'elements' directory.
You're styling a specific button, not the button element.
Also, shouldn't we use a more generic name for the login button style? So that's it's re-usable? `.btn.mod-something` I don't have any ideas that come to mind at the moment, but I think this should at least be less specific.

scss/controls/splash.scss
wfortin: Good point, i'm fixing this.

scss/elements/login-buttons.scss
```diff
@@ -0,0 +1,41 @@
...
+  margin: 0 auto 13px;
+
+  color: $dark-grey;
+  font-size: 14px;
+  font-weight: 400;
+
+  background-color: $white;
+  border: $btn-login-padding-border;
+
+  .icon {
```
marilynechan: Instead of styling the icon directly, could we have something more reusable as Bootstrap `.btn-prepend` for example, that would contain the icon?

scss/elements/color-dot.scss
wfortin: This one's trickier because it's from the StatusPage call, i guess i could map the statuses differently, i'll check that.

scss/utility/layout.scss
```diff
@@ -0,0 +1,54 @@
...
+
+.left {
+  float: left;
+}
+.right {
+  float: right;
+}
+
+.fullscreen {
+  min-height: 100vh;
```
marilynechan: If it's fullscreen, shouldn't it have `min-width: 100vw;` too?

scss/elements/icons.scss
wfortin: According to mockups, it was 16x16 but to achieve pixel perfect centering, 17x17 was easier... What should we do about this?

scss/elements/color-dot.scss
marilynechan: We shouldn't limit ourselves because of the API we use! So +1 for the map. And I guess we'll reuse them elsewhere in the application, for other statuses?

marilynechan:
Great! I like the utility classes you added. We're heading somewhere. Ship it!

scss/elements/login-buttons.scss
wfortin: btn-login is *completely* different from the base buttons of the lib, so i don't actualy see that as a modifier. You don't need to use `.btn.btn-login`, `btn-login` is used standalone.

scss/elements/login-buttons.scss
wfortin: Good point, i'll rename that.

scss/utility/layout.scss
wfortin: I'll add that.

resources/images/polyart.svg
wfortin: Done.

scss/controls/splash.scss
wfortin: I guess since we already have a specific class for this, it makes sense to use the $color var. Other than that, what's the point of having color variables ? Let's discuss about this on slack.

scss/controls/splash.scss
wfortin: Done.

scss/elements/color-dot.scss
wfortin: Yes good point. I'm fixing before merging.

scss/elements/login-buttons.scss
wfortin: Done.

scss/elements/login-buttons.scss
wfortin: Renamed to `.btn-alternate` to be more generic.

scss/elements/color-dot.scss
wfortin: Done, default color-dot color is `$green`

PR #17, opened 2015-10-30
change gulp task for svg to create a svg json, remove a duplicate svg

Unetitebiere:


wfortin:
If the output is in /dist it should be in the repo as we discussed today.

Other than than, it looks good.

wfortin:
Oh i see the file now, it's just so big that github dosen't display it :stuck_out_tongue: 

https://raw.githubusercontent.com/coveo/styleguide/gulp-task-create-svg-json/dist/svg/CoveoStyleGuideSvg.json

Unetitebiere:
Yep, I tried to remove `xmlns:xlink`with attrs and with the plugin `removeEditorsNSData` but it's alway in the output json.

wfortin:
Remove all /target files. They are not needed are from the old version of the styleguide.

styleguide/index.html
```diff
@@ -0,0 +1,91 @@
+<!DOCTYPE html>
```
wfortin: This file should be removed...

PR #21, opened 2015-11-04
New navigation style in Styleguide

marilynechan:


scss/components/navigation.scss
```diff
@@ -0,0 +1,105 @@
...
+    position: relative;
+    float: left;
+    z-index: 899;
+    height: calc(100vh - #{$header-height});
+    width: $navigation-width;
+    background-color: $navigation-background-color;
+    color: $white;
+
+    .navigation-menu-sections {
+      @include slim-scroll(#444444, #D3D3D3);
```
GermainBergeron: We should use a variable for the colors

GermainBergeron:
LGTM :ship: :it:

JPLachance:
Looks good to me! One simple comment, we might want to reduce z-index a little! 899 it's really high :P

docs/_sass/layout/_overrides.scss
```diff
@@ -0,0 +1,6 @@
+// This file contains overrides to the base Styleguide so they will be displayed correctly in the Styleguide documentation.
+.sg-component-display {
+  .navigation, .navigation-menu {
+    max-height: 300px;
```
JPLachance: As discussed, why not simply add an inline style tag in the /components/03_navigation.html template? It would be simpler.

marilynechan: Done.

marilynechan:
I changed the 899 z-index to 51. I did not see anything weird with this value.

PR #22, opened 2015-11-04
Remove table colors and use palette instead

wfortin:
Applied avocode specs, and removed duplicate or old colors to use the basic color palette.


scss/tables/table-actions.scss
```diff
@@ -1,7 +1,7 @@
 @import '../common/colors';
 @import '../controls/filter-box';
 $coveo-table-actions-container-height: 52px;
-$coveo-table-actions-border-width: 2px;
+$coveo-table-actions-border-width: 1px;
```
marilynechan: Why did you need to change this width in more thant one place? I'm referring to the border-width you changed in fixed-header-table.

scss/tables/table.scss
```diff
@@ -43,14 +43,8 @@ table {
...
 
-      border-top: solid 2px $table-borders;
-      border-bottom: solid 2px $table-borders;
-
-      &.metric {
-        color: $heading-2-color;
-
-        background-color: $metric-table-header-color;
-      }
+      border-top: solid 1px $medium-grey;
```
marilynechan: Again changing from 2px to 1px.

marilynechan:
Nice cleanup! Ship it!

scss/tables/table-actions.scss
wfortin: Good point i'll make a variable $table-border in table.scss

JPLachance:
Look's good. Ship it!

PR #23, opened 2015-11-05
Prevent user select on selected table row

JPLachance:
I added "user-select: none " on selected table row. It prevent weird selections on double click. What do you think?


wfortin:
This is kinda sad, because sometimes you want to actually select text...

JPLachance:
Right, but you can select the text if the row is not the selected one. I checked in GG Drive and you simply cannot select the text of any row. It's not that bad in our app if the user can simply unselect the row to select the text (except the case when he got only one row).

wfortin:
Ah, i did't figure that it was only the selected row, but @GermainBergeron 's solution looks great.

JPLachance:
Drop! We will remove user selection using JavaScript on double click instead.

PR #35, opened 2015-11-19
Add utilities section in docs and change svg for organization-status

Unetitebiere:
-remove text spacing with class caps
+add section for utilities in Typography
+fix icon name


docs/_sass/layout/_styleguide.scss
```diff
@@ -25,6 +25,13 @@
   padding: 1.5em;
 }
 
+//text
+div {
+  .break-word, .truncate {
+    width: 145px;
+  }
```
marilynechan: I would add inline style in the documentation instead, to demonstrate the utility of the class.

docs/_components/typography/03_text-utility.html
```diff
@@ -0,0 +1,70 @@
...
+    <div class="right-align">
+        Text in right-align
+    </div>
+
+    <div class="justify">
+        Text in justify
+    </div>
+
+    <div class="nowarp">
+        Text with nowarp
```
marilynechan: Shouldn't it be `nowrap` ?

docs/_components/typography/03_text-utility.html
```diff
@@ -0,0 +1,70 @@
+---
+type: typography
+title: Utilities
+usage: Change typographic weights, styles, and alignment with these utility styles.
```
marilynechan: *utility classes.

docs/_components/typography/03_text-utility.html
```diff
@@ -0,0 +1,70 @@
...
+    <div class="italic">
+        Text in italic
+    </div>
+
+    <div class="caps">
+        Text in caps
+    </div>
+
+    <div class="left-align">
+        text in left-align
```
marilynechan: Small detail, but should be `Text` with an uppercase T.
Also, it reads weird "Text _in_ left-align". With? With the left-align class?

docs/_components/typography/03_text-utility.html
```diff
@@ -0,0 +1,70 @@
...
+    </div>
+
+    <div class="center">
+        text in center
```
marilynechan: Uppercase T here.

docs/_components/typography/03_text-utility.html
```diff
@@ -0,0 +1,70 @@
...
+    <div class="nowarp">
+        Text with nowarp
+    </div>
+
+    <div class="break-word">
+        Text with break-word
+    </div>
+
+    <div class="truncate">
+        Text with truncate is not cut
```
marilynechan: Hmm, shouldn't this read "is cut"?

docs/_components/typography/03_text-utility.html
wfortin: +1

wfortin:
I don't have much more comments than @marilynechan, apply those then :shipit:

docs/_components/typography/03_text-utility.html
```diff
@@ -0,0 +1,70 @@
...
+    </div>
+
+    <ul class="list-reset">
+        <li>list reset item 1</li>
+        <li>item 2</li>
+        <li>item 3</li>
+    </ul>
+
+    <div class="pull-right">
+        text pull-right
```
marilynechan: Uppercase T.
@wfortin We currently have 4 classes in the styleguide that do this (float right/left). Which should we keep, which should we document?

docs/_components/typography/03_text-utility.html
```diff
@@ -0,0 +1,70 @@
...
+    </div>
+
+    <div class="pull-left">
+        text pull-left
```
marilynechan: Uppercase T.

marilynechan:
Ship it!
One last comment: I'm not sure about the `organization-status` icon name. I know I asked for the change in your other PR, but seeing it here makes me realize it's really specific to our application... Maybe 'status' simply? Depending on what the icon represents...

docs/_components/typography/03_text-utility.html
wfortin: `left` and `right` should be the ones used.  (for float left and right)
Please drop `pull-left` and `pull-right`.

Also: these are not "typograpgy" modifiers, we have `align-left` and `align-right` that align the text.

marilynechan:
Great work BTW!

docs/_components/typography/03_text-utility.html
wfortin: Document `align-left` and `align-right`

Unetitebiere:
@marilynechan we have svg with status-** name, I suggest something like status-circle or status-scan ( because the svg looks like a scan :) ). I think status is a bit too generics

PR #38, opened 2015-11-23
+ Add a mod-border-top class for the table actions container

JPLachance:


docs/_components/tables/03_table-actions.html
```diff
@@ -1,11 +1,11 @@
 ---
 type: tables
 title: Table actions container
-usage: This define how to set a filter box for a table and where to place actions button for each table elements.
+usage: This define how to set a filter box for a table and where to place actions button for each table elements. Add the mod-border-top class when the table is not right under the panel-header.
```
wfortin: This defines*
(small typo)

wfortin:
Github needs an "approve" button.

:+1:

scss/tables/table-actions.scss
```diff
@@ -1,5 +1,6 @@
 @import '../common/colors';
 @import '../controls/filter-box';
+@import 'table';
```
marilynechan: Is that really needed?
Sass import works the same way as Less - 'table' is already imported in `guide.scss` so it will be imported twice because of the import you added...

JPLachance: Removed.

marilynechan:
Nothing else to add, :shipit: !

PR #39, opened 2015-11-23
Align table content with header

wfortin:
Added `mod-align-header` to fit padding with heading padding.

Extract variable in file headers to a single `variables.scss`, this will allow easier use of "piece by piece" files and override our default values easily.


scss/tables/table.scss
```diff
@@ -30,6 +27,18 @@ table {
...
 
+  &.mod-align-header {
+    th, td {
+      &:first-child {
+        padding-left: $header-padding;
+      }
+    }
+
+    tbody tr.selected td:first-child {
+      padding-left: 35px;
```
marilynechan: This number surely comes from `$header-padding - [selected-border-width]`?

marilynechan:
Nice cleanup! :ship: :it:

scss/tables/table.scss
wfortin: Good point, fixed it.

GermainBergeron:
LGTM!

PR #40, opened 2015-11-24
Use coveo/slider instead static files

GermainBergeron:


package.json
```diff
@@ -26,6 +26,7 @@
     "url": "https://github.com/Coveo/styleguide.git"
   },
   "devDependencies": {
+    "coveo-slider": "0.9.4",
```
marilynechan: It's a dev dependencies? This kind of surprises me.

docs/_includes/footer.html
```diff
@@ -54,4 +54,5 @@ <h2 class="footer-heading">{{ site.title }}</h2>
 </footer>
 
 <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/2.1.4/jquery.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/underscore.js/1.8.3/underscore-min.js"></script>
```
marilynechan: Shouldn't we bundle JQuery and Underscore if we need them in the Styleguide? And use these dependencies in `head.html` in the documentation?

marilynechan:
Looks good to me, just a few questions!

package.json
wfortin: +1 should be a dependency imo.

DevDependencies aren't downloaded when you do a `npm install coveo-styleguide`

GermainBergeron: It's a dev dependency because it's bundled in the `CoveoStyleGuide.Dependencies.js` file (see gulpfile for more details). If someone pulls the repo and use the files in the `dist` folder, he won't need the slider

docs/_includes/footer.html
wfortin: I actually asked @GermainBergeron to _not_ bundle it :stuck_out_tongue: . I think it's better to state the "requirements" for the libraries but allow users to bring their own versions. 

This also caused problems with the JSUI providing their own versions of jQuery, globalize, underscore, etc.

Makes sense ?

marilynechan: Yep, makes sense. We probably need to write that somewhere in the documentation.

How can we guarantee compatibility with different versions of those libraries?

wfortin: The readme states the dependencies already : https://github.com/coveo/slider#dependencies
To guarantee compatibility, we don't :stuck_out_tongue_closed_eyes: 
We could put disclaimer : tested with Underscore 1.8.3 and jQuery 2.1.4.

marilynechan: Yeah, the coveo-slider readme, but not the coveo-styleguide readme :smile:! And jQuery was already there, so it's not only for the slider.

wfortin:
:ship: it.

package.json
wfortin: Ah makes more sense now.

PR #47, opened 2015-11-26
+add css for new svg in the table

pochretien:


scss/tables/table.scss
```diff
@@ -61,41 +61,41 @@ table {
...
-              opacity: 0.6;
+            .asc-arrow, .desc-arrow {
+              cursor: pointer;
+              opacity: 0.3;
             }
 
-            + i {
-              margin-top: 2px;
+            .asc-arrow, .desc-arrow {
+              &:hover {
```
marilynechan: Shouldn't this go after line 75? You repeat `.asc-arrow, .desc-arrow` here.

scss/tables/table.scss
```diff
@@ -61,41 +61,41 @@ table {
...
+          display: inline-flex;
 
-          i {
-            position: relative;
-            top: 3px;
-            left: -11px;
-
-            display: block;
+          .coveo-tables-sort {
+            position: absolute;
```
marilynechan: Is it really necessary to position this with absolute?

scss/tables/table.scss
```diff
@@ -61,41 +61,41 @@ table {
...
-          top: 8px;
+          display: inline-flex;
 
-          i {
-            position: relative;
-            top: 3px;
-            left: -11px;
-
-            display: block;
+          .coveo-tables-sort {
```
marilynechan: Avoid the coveo or admin prefix, particularly in the styleguide!

scss/tables/table.scss
pochretien: I need to set absolute if a want to make the sort icon with an height higher than the row. That's not all the header which are the same height

scss/tables/table.scss
```diff
@@ -61,41 +61,39 @@ table {
...
-            left: -11px;
-
-            display: block;
+          .tables-sort {
+            position: absolute;
+            top: 25%;
 
-            opacity: 0.3;
+            width: 23px;
+            height: 16px;
```
marilynechan: Have you tried using standard icon sizing classes (just `icon` or with `mod-lg` modifier)?
There should be no need for absolute positioning if you display the admin-sort-icon inline.

wfortin: +1

scss/tables/fixed-header-table.scss
```diff
@@ -1,7 +1,7 @@
 .fixed-header-table-container {
   position: relative;
 
-  padding-top: 27px;
+  padding-top: 32px;
```
marilynechan: You should have a variable in `variables.scss` for this, you use this value at more than 1 place.

scss/variables.scss
```diff
@@ -13,7 +13,7 @@ $filter-box-height: 32px;
 // Header
 $header-height: 65px;
 $header-application-offset-y: 6px;
-$header-padding: 40px;
+$header-padding: 35px;
```
marilynechan: AFAIK, this didn't change in the mockups. It should stay 40px.

scss/tables/table.scss
```diff
@@ -146,6 +136,10 @@ table {
 
       border-bottom: $table-border;
 
+      &:first-child {
+        border-left: $table-selected-border-width solid transparent;
```
marilynechan: Nice, way better :-) .

marilynechan:
Looks good to me, ship it!

wfortin:
:shipit: GO

PR #48, opened 2015-11-27
Add margin-left and right auto white-space utilities.

wfortin:


marilynechan:
Curiosity, in which case is this useful?

Ship it!

wfortin:
Useful for flex elements (makes kind of a float)

pochretien:
:shipit:

PR #52, opened 2015-12-03
* Add dropdown max height

vseguin:


scss/controls/dropdown.scss
```diff
@@ -70,5 +70,7 @@
 
   &.open > .dropdown-menu {
     display: block;
+    max-height: 400px;
+    overflow-y: scroll;
```
marilynechan: I would add these properties directly in `.dropdown-menu` (not with the open state). Also, use `auto` instead of `scroll`! My bad.

marilynechan:
Apply review then ship it!

PR #54, opened 2015-12-04
- Remove specific color from toggle-label in toggle widget

vseguin:
It should be blue as the other text.


wfortin:
:+1:

wfortin:
Did you run `gulp sass:format` before submitting ?

pastjean:
:+1:

PR #55, opened 2015-12-08
Set table header font for 12px, sort icon size and filter icon size

pochretien:
fix filter icon position with the new size with mod-lg
fix table sort icon size and set table header size for 12px

### DIFF
--- a/scss/tables/table.scss
+++ b/scss/tables/table.scss
@@ -60,9 +60,10 @@ table {
           display: inline-flex;
 
           .tables-sort {
-            vertical-align: middle;
             width: 15px;
             height: 15px;
+
+            vertical-align: middle;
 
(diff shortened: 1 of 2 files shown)


scss/tables/table.scss
```diff
@@ -39,7 +39,7 @@ table {
     height: $table-header-height;
 
     color: $medium-blue;
-    font-size: 15px;
+    font-size: 12px;
```
marilynechan: Just making sure - will you update the big-table font-size in the admin? We want all table headers to have the same font-size. Don't mind my comment if there's no specific font-size in the big-table :-) .

marilynechan:
If the build passes, ship it!

wfortin:
:+1:

PR #56, opened 2015-12-08
Minor fixes

JPLachance:
* Move the prompt style to it's own file
- Remove the btn right margin, replace it by a margin between two buttons

### DIFF
--- a/scss/forms/table-form.scss
+++ b/scss/forms/table-form.scss
@@ -123,7 +123,7 @@ form {
           .dropdown-toggle {
             box-sizing: border-box;
             width: 100%;
-            padding-right: 11px;
+            padding-right: 6px;
             padding-left: 0;
 
             font-size: 16px;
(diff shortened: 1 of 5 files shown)


scss/forms/table-form.scss
```diff
@@ -123,7 +123,7 @@ form {
           .dropdown-toggle {
             box-sizing: border-box;
             width: 100%;
-            padding-right: 11px;
+            padding-right: 6px;
```
marilynechan: AFAIK, I changed this so the arrow would be aligned with the other inputs icon.

JPLachance: I removed the `margin-right: 5px;` on the btn class which was causing your alignment issue. Yes it's aligned correctly ;)

marilynechan: Great, just making sure ;-) .

pochretien:
:shipit:

PR #58, opened 2015-12-10
Feature tooltips

marilynechan:
* New tooltips styling, with Bootstrap plugin, documented.
* Small modification to the flex-form (removed negative margins hack).

### DIFF
--- a/docs/_components/components/04_tooltip.html
+++ b/docs/_components/components/04_tooltip.html
@@ -1,16 +1,14 @@
 ---
 type: component
 title: Tooltip
-usage: Use to display additional information.
+usage: Use to display a tooltip on any element. We use Bootstrap Tooltip plugin.
 stylesheet: '/scss/components/tooltip.scss'
 ---
-<!-- Note that positioning is automatically done by Bootstrap library.
-     There's no need to put a tooltip in a relative positioned element, nor
-     to put inline style on the tooltip. -->
-<div class="relative">
-    <button class="btn mod-primary">Do something!</button>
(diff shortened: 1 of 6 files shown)


docs/_components/components/04_tooltip.html
```diff
@@ -0,0 +1,14 @@
...
+type: component
+title: Tooltip
+usage: Use to display a tooltip on any element. We use Bootstrap Tooltip plugin.
+stylesheet: '/scss/components/tooltip.scss'
+---
+<button class="btn mod-primary" title="Some helpful information!" data-placement="right">Hover me</button>
+<div class="m2">
+    <div class="inline-block" title="Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore
+        et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat."
+         data-placement="bottom" data-trigger="click">Another sample, click me!
```
wfortin: Oh nice ! data-trigger

wfortin: Could be interesting to leverage this technique and only use bootstrap tooltip.js instead (or maybe : [popover.js](https://github.com/twbs/bootstrap/blob/master/js/popover.js)) instead of also having drop.js for sources error popup.

wfortin:
Great job! :+1:

pochretien:
Nice!! :ship: it

PR #60, opened 2015-12-10
Fix add -state-stopped color dot

pochretien:


scss/elements/color-dot.scss
```diff
@@ -4,7 +4,7 @@
   background-color: $green;
   border-radius: 100%;
 
-  &.state-critical {
+  &.state-critical, &.state-stopped {
```
marilynechan: Can't you map your status to `.state-critical` in the admin app instead of creating a new state?

PR #61, opened 2015-12-11
fix and clean some svg

pochretien:


wfortin:
xml version is not needed. Other than that :shipit:

marilynechan:
Ship it!

PR #70, opened 2015-12-15
Fix alignments

marilynechan:
* Flex form
* Margin between butons
* Dropdown alignment.


wfortin:
:+1:

scss/forms/flex-form.scss
```diff
@@ -2,15 +2,22 @@
   flex-flow: row wrap;
 
   display: flex;
+  margin-top: -10px;
+  margin-left: -10px;
```
pochretien: its a good idea to use a variable for this margin ? it seem the same margin for the class below but im not sure its for the same reason

pochretien:
:ship: it

scss/forms/flex-form.scss
marilynechan: Good idea, I added a variable!

PR #71, opened 2015-12-16
!Fix actionbar dropdown clipping and ie width

frederik-labbe:
[ADUI-352]

### DIFF
--- a/scss/components/application-container.scss
+++ b/scss/components/application-container.scss
@@ -2,6 +2,6 @@
   margin-top: $header-height;
   margin-left: $navigation-width;
   overflow: hidden;
-
+  min-height: 600px;
   border-top-left-radius: 2px;
 }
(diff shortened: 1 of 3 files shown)


marilynechan:
Don't forget to run `gulp sass:format` before submitting your PR, that's why the build failed.
Ship it!

scss/components/application-container.scss
```diff
@@ -2,6 +2,6 @@
   margin-top: $header-height;
   margin-left: $navigation-width;
   overflow: hidden;
-
+  min-height: 600px;
```
pochretien: I think we can add this 600px in a variable because we use it for the application container

pochretien:
sass format and :ship:  it

wfortin:
:+1:

PR #74, opened 2015-12-18
! Fix collapsible z-index

wfortin:
I guess z-index was there for a reason? But it doesn't show up in the license panel and removing it still works in all sources panels...
Let me know if you know why there was a z-index: -1;


pastjean:
:shipit:

JPLachance:
No clue... ship it!

PR #78, opened 2015-12-24
! Fix flex-auto breaking flex footer scroll

wfortin:


wfortin:
https://github.com/basscss/flex-object/blob/master/lib/flex.css

Looks like the fix suggested by BassCSS is not valid...

GermainBergeron:
:shipit:

JPLachance:
:shaved_ice: :icecream:

PR #79, opened 2016-01-06
fix glitch with dropdown menu on table action

pochretien:
remove display flex glitch
fix position for the more dropdown


JPLachance:
Look's good!

wfortin:
LGTM

PR #90, opened 2016-01-28
! Fix tables background colors

JPLachance:


scss/tables/table.scss
```diff
@@ -291,7 +286,9 @@ table {
     }
 
     &.selected {
-      background-color: $block-background-color;
+      &:not(:hover) {
```
wfortin: :not?

Can you explain

JPLachance: The background-color for the hover state has priority over the background-color of the selected state.

PR #94, opened 2016-02-10
Remove font-weight:300 on small font sizes for linux/IE

wfortin:
[ADUI-675]

### DIFF
--- a/scss/components/member.scss
+++ b/scss/components/member.scss
@@ -33,7 +33,7 @@
       margin-top: 2px;
 
       font-size: 12px;
-      font-weight: 300;
+      font-weight: 400;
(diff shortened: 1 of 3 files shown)


JPLachance:
:ship: :it:

scss/components/member.scss
```diff
@@ -33,7 +33,7 @@
       margin-top: 2px;
 
       font-size: 12px;
-      font-weight: 300;
+      font-weight: 400;
```
GermainBergeron: For the explanation https://developer.mozilla.org/en/docs/Web/CSS/font-weight#Fallback

GermainBergeron:
:shipit:

PR #101, opened 2016-03-01
! Fix font colors and padding

wfortin:
* Use text-color for darker text
* Clean up padding conflict with table style rules
image

### DIFF
--- a/scss/lib/datepicker.scss
+++ b/scss/lib/datepicker.scss
@@ -219,9 +219,9 @@ tbody.datepickerDays td.datepickerSelected {
   background-color: #0077cc;
 }
 
-/* Grey out dates in the future */
-div.datepicker td.datepickerFuture a {
-  color: #dddddd;
+/* yellow cell background on hover */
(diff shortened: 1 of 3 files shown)


JPLachance:
Look's good!

scss/lib/datepicker.scss
```diff
@@ -219,9 +219,9 @@ tbody.datepickerDays td.datepickerSelected {
   background-color: #0077cc;
 }
 
-/* Grey out dates in the future */
-div.datepicker td.datepickerFuture a {
-  color: #dddddd;
+/* yellow cell background on hover */
+div.datepicker tbody.datepickerDays td:hover {
+  background-color: #ffcc33;
```
pochretien: can extract this color in variable or its only used for this datepicker ?

pochretien:
:shipit:

scss/lib/datepicker.scss
wfortin: It's the base libraries colors, we could clean up the unused classes someday.

PR #102, opened 2016-03-02
New blankslate component

wfortin:
Small demo http://codepen.io/wfortin/pen/yONMMK?editors=1100

image


GermainBergeron:
:shipit:

scss/typography/headings.scss
```diff
@@ -1,6 +1,6 @@
 h1, h2, h3, h4, h5, h6 {
+  color: $medium-blue;
```
JPLachance: I'm not so sure it's a good idea to make all titles always blue by default. In the new modal, titles are white and we are gonna have overrides. I prefer to use "color classes" where we need them.

wfortin: Good point this should not have made it to the PR, I was experimenting.

Removing it.

JPLachance:
:ship: :it: !

PR #104, opened 2016-03-03
Chosen custom single select

pochretien:
+add chosen lib
+add chosen exemple in docs
+add custom css for the chosen single select

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "coveo-slider": "0.9.8"
   },
   "devDependencies": {
+    "chosen-npm": "1.4.2",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -30,6 +30,7 @@
     "coveo-slider": "0.9.8"
   },
   "devDependencies": {
+    "chosen-npm": "1.4.2",
```
wfortin: This should be a dependency, not a devDependency

wfortin:
Nice !!

GermainBergeron:
:shipit:

PR #107, opened 2016-03-14
Implement the flat select mod-btn-group styling

JPLachance:


scss/controls/flat-select.scss
```diff
@@ -64,6 +67,32 @@
...
+    height: $button-height;
+    padding: $button-padding-y $button-padding-x;
+
+    font-size: $button-font-size;
+    line-height: $button-line-height;
+
+    border-radius: 0;
+
+    &:first-child {
+      border-radius: 2px 0 0 2px;
```
wfortin: We might want to extract the border radius to a variable.

docs/_components/controls/15_flat-select-btn-group.html
```diff
@@ -0,0 +1,14 @@
...
+usage: Modifiers for the Flat select control { mod-btn-group } .
+stylesheet: '/scss/controls/flat-select.scss'
+---
+
+<div class="flat-select mod-btn-group">
+    <a class="flat-select-option selectable">Option 1</a>
+    <a class="flat-select-option">Option 2</a>
+    <a class="flat-select-option selectable caps bold">Option 3</a>
+    <a class="flat-select-option selectable caps bold">Option 4</a>
+    <a class="flat-select-option caps bold">Option 5</a>
```
wfortin: What does `selectable` do? If it's absent, you can't select it ?

JPLachance: Exactly, it's legacy from our actual flat select.

scss/controls/flat-select.scss
JPLachance: I'm not so sure that's required since value is different for both first-child and last-child.

wfortin: The `2px` value should be in a var though.

wfortin:
Extract to radius to a var then :shipit: !

scss/controls/flat-select.scss
```diff
@@ -17,10 +16,20 @@
 
     background-color: $medium-blue;
     border: 1px solid $medium-blue;
-    border-radius: 2px;
+    border-radius: $border-radius;
 
     cursor: default;
 
+    .icon-container .icon {
+      vertical-align: -0.2em;
```
GermainBergeron: :confused: Can you avoid this?

GermainBergeron:
:ship: :it:

pochretien:
GG. :ship: :it:

scss/controls/flat-select.scss
JPLachance: Nope. I tried, but the default alignment values of the Styleguide for icons are not working when using icons mod-lg. I discussed with William and we won't change the default value now because we may break everything. But, I could try to align icons using flex..

PR #108, opened 2016-03-14
add style when we want to delete a row

pochretien:
add red color with opacity for the selected row to be deleted
add padding for the button in prompt-action
fix the css for the prompt action inline

### DIFF
--- a/scss/tables/table-actions.scss
+++ b/scss/tables/table-actions.scss
@@ -38,6 +38,10 @@
         color: $medium-blue;
         font-weight: bold;
       }
+
+      &.prompt-action {
+        padding: 8.5px 16px;
+      }
     }
   }
 
--- a/scss/tables/table-actions.scss
+++ b/scss/tables/table-actions.scss
@@ -40,7 +40,7 @@
       }
 
       &.prompt-action {
-        padding: 8.5px 16px;
+        padding: $button-padding-y $button-padding-x;
       }
     }
   }
--- a/scss/tables/table.scss
+++ b/scss/tables/table.scss
@@ -308,10 +308,10 @@ table {
     }
   }
 
-  &.mod-admin-table-view-delete {
+  &.mod-deleting {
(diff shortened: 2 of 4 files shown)


scss/tables/table.scss
```diff
@@ -307,4 +307,12 @@ table {
       transition: background-color 300ms;
     }
   }
+
+  &.mod-admin-table-view-delete {
+    tbody {
+      tr.selected {
+        background-color: rgba(244, 67, 54, 0.1);
```
GermainBergeron: variable for color?

scss/tables/table-actions.scss
```diff
@@ -38,6 +38,10 @@
         color: $medium-blue;
         font-weight: bold;
       }
+
+      &.prompt-action {
+        padding: 8.5px 16px;
```
GermainBergeron: half pixel aren't recommended

GermainBergeron:
Fix it, then :shipit:

scss/tables/table-actions.scss
pochretien: its a copy past forme the btn class, I think too its not a good idea but i have not choice for now if i want the same size than the button beside him

scss/tables/table.scss
pochretien: ya, I check if I can do the same think with hex variable because for now, the rgb color its the same like red variable

scss/tables/table-actions.scss
JPLachance: Plz use button variables (padding: $button-padding-y $button-padding-x;).

scss/tables/table.scss
JPLachance: We are not suppose to use any other color than the ones in the palette. If the mockup is not using a valid color, ask Marjorie.

JPLachance:
Ship it!

PR #119, opened 2016-03-23
Tweaks for the textbox!

JPLachance:


docs/_components/form_controls/01_input.html
```diff
@@ -1,10 +1,27 @@
...
+    <div class="textbox form-group">
+        <input class="textbox-input" type="text" value="My valid value" required disabled/>
+        <label class="textbox-label">Disabled valid input</label>
+    </div>
+
+    <div class="textbox form-group">
+        <input class="textbox-input" type="text" required disabled/>
+        <label class="textbox-label">Disabled invalid input</label>
+    </div>
 </div>
```
GermainBergeron: Should we add the script here?

PR #121, opened 2016-03-24
Textbox validation feedback!

JPLachance:
* Move the textbox style into his own file
+ Add style for the valid/invalid textbox


wfortin:
As we discussed, let's name that `input-field` as our friends at Materialize do.

scss/controls/input-field.scss
```diff
@@ -0,0 +1,120 @@
+.input-field {
+  position: relative;
+
+  max-width: $input-field-max-width;
```
GermainBergeron: I'm not sure about the max-width, IMO it should take 100%

scss/controls/input-field.scss
```diff
@@ -0,0 +1,120 @@
...
+  margin-top: $input-field-margin-top;
+
+  font-size: 15px;
+
+  input {
+    width: 100%;
+    height: 34px;
```
GermainBergeron: var for this?

GermainBergeron:
Try to use variables when possible, some numbers looks magic to me

scss/controls/input-field.scss
wfortin: We discussed about this, and until we don't have a grid, we capped the input width to a fixed width.

Makes sense ?

scss/controls/input-field.scss
wfortin: there is already: `$filter-box-height`, and `$button-height`. We should consolidate this to a single variable.

JPLachance: Actually, we are respecting the UX team specification, and yes it's inconsistant. I'll ask them what we can do about that.

JPLachance: I don't think we should merge all those variables into one. My point is that if you change that value, it will probably cause bugs to one or another of those inputs. I've added a variable for the input-field height.

PR #122, opened 2016-03-29
Fixes for the table header issues

JPLachance:
! Fix the table actions height for IE
! Fix the mod-collapsible-rows scope (fix the header row height


pochretien:
:shipit:

PR #127, opened 2016-04-05
Feature new progress bar

pochretien:


docs/_components/form_controls/18_progress-bar.html
```diff
@@ -0,0 +1,30 @@
+---
+type: formControls
+title: Progress bar
+usage: Show a progression between 0 to 100%.
```
GermainBergeron: Show a progression between 0 and 100%.

docs/_components/form_controls/18_progress-bar.html
```diff
@@ -0,0 +1,30 @@
...
+stylesheet: '/scss/elements/progress-bar.scss'
+---
+
+<div class="bg-medium-blue progress-bar-exmple-container">
```
GermainBergeron: typo in "example" (missing a)

docs/_components/form_controls/18_progress-bar.html
```diff
@@ -0,0 +1,30 @@
...
+    <div class='progress progress-40'>
+      <span class='progress-bar-completed'></span>
+    </div>
+  </div>
+</div>
+
+<button class="btn mod-primary js-exemple-progress mt2">Test progression</button>
+
+<script>
+  $('.js-exemple-progress').click(function(event){
```
GermainBergeron: class mismatch for the selector

GermainBergeron: exemple => example

scss/elements/progress-bar.scss
```diff
@@ -0,0 +1,38 @@
+.progress-bar {
+  width: 100%;
+
+  text-align: center;
+}
+.progress {
+  position: relative;
+  z-index: 1;
+
+  height: 20px;
```
GermainBergeron: Maybe you should use a couple of variable for the height, margin & border-radius

GermainBergeron:
:shipit:

scss/elements/progress-bar.scss
```diff
@@ -0,0 +1,38 @@
+.progress-bar {
+  width: 100%;
+
+  text-align: center;
+}
+.progress {
```
JPLachance: Is there any reason why your code isn't scoped under the `.progress-bar` class? I think it should be.

JPLachance:
Scope your style then :shipit: !

scss/elements/progress-bar.scss
pochretien: only an habitude to create flat css but I can scope

wfortin: I would scope also

scss/elements/progress-bar.scss
```diff
@@ -0,0 +1,38 @@
...
+  overflow: hidden;
+
+  background-color: $white;
+}
+
+@for $i from 0 through 100 {
+  .progress-#{$i} .progress-bar-completed {
+    width: percentage($i / 100);
+  }
+}
```
wfortin: This outputs a 100 classes, it might be a little heavy, we could use inline style to set the width ? @GermainBergeron @JPLachance what do you think?

GermainBergeron: 100 classes is a lot, I think you're right and we should use inline style (this would also allows us to use more precision) 
I'd create a small script and put it under the tag in the styleguide

PR #130, opened 2016-04-12
Input tweaks

JPLachance:
* Handle inline-help-tooltip in labels
* Add coveo-checkbox-labels-group for checkboxes group
* Add word-wrap: break-word for tooltips with a long text
* No more bold for inputs label

### DIFF
--- a/docs/_components/form_controls/01_input.html
+++ b/docs/_components/form_controls/01_input.html
@@ -12,7 +12,20 @@ stylesheet: '/scss/elements/input-field.scss'
 
     <div class="input-field form-group">
         <input type="text" required placeholder="Placeholder text"/>
-        <label>Enabled invalid input</label>
+        <label>
+            Enabled invalid input
+            <span class="inline-help-tooltip" title="A simple inline help text." data-placement="right">
+                <span class="inline-help-icon">
+                    <svg viewBox="0 0 20 20" class="icon fill-medium-grey">
+                        <path fill-rule="evenodd" clip-rule="evenodd"
+                              d="M10 4.6c-2 0-3.7 1.6-3.7 3.7h1.8c0-1 .8-1.8 1.8-1.8s1.8.8 1.8 1.8c0 1.8-2.8 1.6-2.8 4.6h1.8c0-2.1 2.8-2.3 2.8-4.6.2-2.1-1.5-3.7-3.5-3.7z"></path>
+                        <path d="M9.1 13.8h1.8v1.8H9.1z"></path>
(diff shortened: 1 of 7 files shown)


docs/_components/form_controls/01_input.html
```diff
@@ -12,7 +12,20 @@
...
+            Enabled invalid input
+            <span class="inline-help-tooltip" title="A simple inline help text." data-placement="right">
+                <span class="inline-help-icon">
+                    <svg viewBox="0 0 20 20" class="icon fill-medium-grey">
+                        <path fill-rule="evenodd" clip-rule="evenodd"
+                              d="M10 4.6c-2 0-3.7 1.6-3.7 3.7h1.8c0-1 .8-1.8 1.8-1.8s1.8.8 1.8 1.8c0 1.8-2.8 1.6-2.8 4.6h1.8c0-2.1 2.8-2.3 2.8-4.6.2-2.1-1.5-3.7-3.5-3.7z"></path>
+                        <path d="M9.1 13.8h1.8v1.8H9.1z"></path>
+                        <path d="M10 17.2c-4 0-7.2-3.2-7.2-7.2S6 2.8 10 2.8 17.2 6 17.2 10 14 17.2 10 17.2zM10 1c-5 0-9 4-9 9s4 9 9 9 9-4 9-9-4-9-9-9z"
+                              clip-rule="evenodd" fill-rule="evenodd"></path>
+                    </svg>
```
GermainBergeron: Can we use something like you did for the link to the stylesheet?

scss/components/tooltip.scss
```diff
@@ -112,6 +112,7 @@
 
     color: $pure-white;
     text-decoration: none;
+    word-wrap: break-word;
```
GermainBergeron: Should we put an hyphen when we break words?

JPLachance: In my specific case, i'm breaking an url.. not so sure it's a good idea :)

GermainBergeron: This code will break everything that takes more than one line, not just an url

JPLachance: Tooltip max length is 300px. Without the word-wrap, everything that is more than 300px long is breaking the tooltip. I think we should keep it and it one day it causes a problem somewhere, I'll remove it.

docs/_components/form_controls/01_input.html
JPLachance: {{ site.data.icons['help'] }} woks! But I had to add the fill: $medium-grey; for that specific icon in the sass code.

PR #138, opened 2016-04-19
Added Fork in CodePen button

GermainBergeron:
Moved content from scripts.html to a scripts.js file
Added a task to copy js files
Added a script to generate the fork in CodePen


resources/js/scripts.js
```diff
@@ -0,0 +1,128 @@
+$(document).ready(function() {
```
GermainBergeron: Moved from scripts.html

PR #142, opened 2016-04-27
Feature Panel Header

GermainBergeron:
Add panel header script
Apply new mockups

image


wfortin:
Awesome. You'll probably need to cleanup unused sass in v7 and v8, make sure you do!

scss/components/panel-header.scss
```diff
@@ -0,0 +1,42 @@
...
+
+  border-bottom: solid 1px $medium-grey;
+
+  .action-bar {
+    flex-flow: row nowrap;
+
+    display: flex;
+  }
+
+  .action {
```
JPLachance: Isn't that weird to have two class named "action" defined at two different places in almost the same context? See table-actions.scss.

GermainBergeron: In fact, in the admin we re-use the same view (ActionBarView), so the markup is the same 😉

PR #150, opened 2016-05-16
* Aligned filter icon with text

sroy3:
screen shot 2016-05-16 at 8 44 08 am

instead of

screen shot 2016-05-16 at 8 44 25 am


JPLachance:
Nice! Just one question. What if we change our current approche for aligning those icons by something a little more sophisticated? Like flex? It's a painful move, but I think it would be better.

PR #152, opened 2016-05-17
add a new mode full content

pochretien:
fix style for modal and prompt in docs
add new prompt info

### DIFF
--- a/scss/utility/layout.scss
+++ b/scss/utility/layout.scss
@@ -57,6 +57,11 @@
   min-height: 100vh;
 }
 
+.full-content {
+  width: 100%;
+  height: 100%;
+}
+
 .hidden {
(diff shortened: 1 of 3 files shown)


scss/utility/layout.scss
```diff
@@ -57,6 +57,11 @@
   min-height: 100vh;
 }
 
+.full-content {
```
JPLachance: I'm not convinced by that class name. Can we think about something a little better?

JPLachance:
Great job, thanks for that re-work!

scss/utility/layout.scss
pochretien: I tried to find a quick name but relevant to set the content to max space, do you have suggestion ?

PR #156, opened 2016-05-19
+ Add missing border utils classes

JPLachance:
* Update scss-lint and gulp-csscomb

Cannot generate those classes using a loop at the moment since there is a bug in gonzales-pe used by csscomb.


wfortin:
:+1:

pochretien:
👍

PR #160, opened 2016-05-24
add check for missing files for dependencies

pochretien:


gulpTasks/lib.js
```diff
@@ -1,6 +1,8 @@
 var config = require('./configuration');
 
 var gulp = require('gulp-help')(require('gulp'));
+var _ = require('underscore');
```
JPLachance: Is it used?

JPLachance:
Good job!

gulpTasks/lib.js
pochretien: nope, im doing test because travis dont like Strings.prototype function
removed

PR #163, opened 2016-05-27
Added style when an li is active

PMarcL:


scss/controls/dropdown.scss
```diff
@@ -75,6 +75,13 @@
         }
       }
 
+      &.active {
+        color: $dark-blue;
+
+        background: none;
```
sroy3: You could merge these as `background: $light-grey;`

PMarcL: yep, I'll do that

wfortin:
You need to run `gulp sass:format` to format the sass to our standards.

PR #164, opened 2016-05-30
Lightbox autocomplete fix

PMarcL:
The autocomplete-value selector was outside the scope of the lightbox-search selector which was problematic in a few table-forms.


GermainBergeron:
:shipit:

PR #165, opened 2016-05-31
Inputs validation tweaks

JPLachance:
* Improve the message-alert svg
* Tweaks for the validation error
- Remove duplicated code in cover-table-form

### DIFF
--- a/resources/icons/svg/message-alert.svg
+++ b/resources/icons/svg/message-alert.svg
@@ -1,8 +1,7 @@
-<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
-	 width="512px" height="512px" viewBox="0 0 512 512" enable-background="new 0 0 512 512" xml:space="preserve">
-    <circle fill="black" stroke="black" stroke-width="20" stroke-dasharray="null" stroke-linejoin="null" stroke-linecap="null" cx="256" cy="256" r="246" class="circle"/>
-  <g class="exclamation">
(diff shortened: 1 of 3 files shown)


resources/icons/svg/message-alert.svg
```diff
@@ -1,8 +1,7 @@
-<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
-	 width="512px" height="512px" viewBox="0 0 512 512" enable-background="new 0 0 512 512" xml:space="preserve">
-    <circle fill="black" stroke="black" stroke-width="20" stroke-dasharray="null" stroke-linejoin="null" stroke-linecap="null" cx="256" cy="256" r="246" class="circle"/>
-  <g class="exclamation">
-   <rect id="svg_3" height="193" width="65" y="95" x="224" stroke-linecap="null" stroke-linejoin="null" stroke-dasharray="null" stroke-width="null" fill="white"/>
-   <rect id="svg_4" height="67" width="63" y="350" x="224" stroke-linecap="null" stroke-linejoin="null" stroke-dasharray="null" stroke-width="null" fill="white"/>
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512" enable-background="new 0 0 512 512" fill="white">
```
pochretien: not sure about the fill white here, i think we can use the css class fill-white

JPLachance: It works perfectly as is and that way our icon is not depending on our css. Otherwise, when changing the css we would break icons without knowing why.

PR #167, opened 2016-06-02
add mod medium size for the prompt

pochretien:


scss/components/prompt.scss
```diff
@@ -31,4 +31,15 @@
       }
     }
   }
+
+  &.mod-medium {
```
JPLachance: What's the point of adding that if it does nothing?

pochretien: it set the max-height and min with the previous size of the mod-medium in modal.scss if we use mod-prompt with mod-medium

wfortin:
:shipit:

PR #171, opened 2016-06-08
Updated the dropdown form control

sroy3:
screen shot 2016-06-08 at 11 17 11 am

### DIFF
--- a/scss/common/palette.scss
+++ b/scss/common/palette.scss
@@ -4,7 +4,6 @@ $dark-blue: #263e55;
 $medium-blue: #1d4f76;
 $light-blue: #296896;
 $dark-grey: #373737;
-$grey: #999999;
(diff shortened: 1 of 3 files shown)


scss/common/palette.scss
```diff
@@ -4,6 +4,7 @@ $dark-blue: #263e55;
 $medium-blue: #1d4f76;
 $light-blue: #296896;
 $dark-grey: #373737;
+$grey: #999999;
```
JPLachance: That grey is not in our product guideline: http://styleguide.coveo.com/d/GthysWU8RY0Q/product-guidelines
I would use $medium-grey instead.

sroy3: I tried it and felt it was too light, but not too light not to use it. I will change it.

PR #182, opened 2016-06-21
Cleanup unused colors reported on css-stats

wfortin:
AW YIS!

According to http://cssstats.com/stats?link=http%3A%2F%2Fcoveo.github.io%2Fvapor%2Fdist%2Fcss%2FCoveoStyleGuide.css

I reviewed and removed/updated the colors.

Ditched commons/colors, we'll now only use palette.scss


JPLachance:
Plz bring back the stratos color :P

scss/tables/table.scss
```diff
@@ -306,8 +306,8 @@ table {
 
   &.confirm-delete {
     .row-selected td, &.row-selected:hover td {
-      background-color: $invalid-background-color;
-      border-color: $invalid-message-background-color;
+      background-color: $white;
```
JPLachance: Hum, nope. $invalid-background-color was a light red, and you changed it for $white (a light gray). Not sure we want to change that.

JPLachance: You may set $red with an opacity of like 20%.

wfortin: I'll go with that.

GermainBergeron:
:shipit:

PR #183, opened 2016-06-22
Add error style for non standard input elements

rousselm4:


wfortin:
LGTM :shipit:

scss/forms/form-validation.scss
```diff
@@ -42,6 +42,14 @@
   }
 }
 
+.generic-form-error {
+  display: inline-block;
+  margin-bottom: 10px;
+
+  color: $red;
+  font-size: 12px;
```
pochretien: why we use a font-size 12 here ?

rousselm4: It's the size used in the same file for similar errors.

PR #184, opened 2016-06-22
Dropdown arrow and aligment fix

JPLachance:
+ Add a svgs.scss file containing inlined svgs
* Make them customizable
! Fix the dropdown arrow vertical alignment in table-form
! Fix the dropdown arrow when dropdown is opened
+ Add a rule to the reset.scss file to remove the default button margin in Safari


wfortin:
Nice job! But in my PR i got rid of `$stratos` as it wasn't used, but i can bring it back....

PR #188, opened 2016-06-27
Removed max-height on dropdown-menu when in table-actions-container

sroy3:


scss/tables/table-actions.scss
```diff
@@ -83,6 +83,10 @@
         margin-left: $button-margin-width;
       }
     }
+
+    .dropdown-menu {
```
JPLachance: You could insert it above, line 57, since there is already custom code for the dropdown in the action bar there ;)

sroy3: Is it worth wrapping the element in more inheritance? It is already 3 levels deep.

wfortin: I'd go on @sroy3 's side. We should try to not nest the class as much as possible.

JPLachance: Personally, I prefer when a modification on the original style is scoped precisely to prevent side effects. And, it looks deep, but when you look at it, a dropdown-menu cannot be anywhere else than in a dropdown..

wfortin: After discussing with @JPLachance I think it would be better to scope on the "More" dropdown, since it's the only one that we want to affect. The predicate filters should have the default max-height.

I think that even better, we'd have a modifier on the dropdown that would set max-height none. And therefore, we wouldn't need to scope it!

sroy3: I have added a modifier and since the max-height was on `.dropdown .dropdown-menu` I moved `.dropdown-menu` out.

scss/controls/dropdown.scss
```diff
@@ -2,129 +2,129 @@
   position: relative;
 
   display: inline-block;
+}
 
-  .dropdown-prepend {
-    padding-right: 5px;
+.dropdown-prepend {
```
wfortin: Oouch... not sure about this mix.
All classes refering to the dropdown should be scoped within the `.dropdown` class. `.dropdown-toggle` for example is meant to be used within a `.dropdown` so we should nest it as it was before.

sroy3: All elements that were nested in the `.dropdown` class already have a `.dropdown-` prefix clearly identifying them to the dropdown. It is a useless nesting that restricts us from using simple modifiers without nesting those modifiers or using `!important` which is not very clean.

wfortin: Yeah, that's obvious now!

Let's try to remember that for other components. I like that actually! Let's go with this solution.

PR #193, opened 2016-07-07
Fixed on hover for table rows colors for IE

PMarcL:


scss/tables/table.scss
```diff
@@ -271,6 +271,11 @@ table {
 
     &:hover {
       background-color: $light-grey;
+
+      td {
+        // ie fix
+        background-color: $light-grey;
```
sroy3: Instead of also adding the color on the td element and specifying that it is a fix for IE, I would simply add the color only on the td element always.

```
&:hover td {
    background-color: $light-grey;
}
```

PMarcL: done

PR #194, opened 2016-07-13
Added a variable, changed links color

PMarcL:


wfortin:
Add the color to the documentation also please.

PMarcL:
@wfortin Since this color is only used for links and is not in our palette and it's not used in any class, I don't think I should add this to the documentation except if I add a link section somewhere, but it wouldn't be very useful.

wfortin:
Yeah i guess you're right. :shipit:

PR #198, opened 2016-07-19
+ Added a facet component with its style

sroy3:
screen shot 2016-07-19 at 9 45 01 am

### DIFF
--- a/docs/_components/components/10_facet.html
+++ b/docs/_components/components/10_facet.html
@@ -13,7 +13,7 @@ stylesheet: '/scss/components/facet.scss'
                         <svg viewBox="0 0 10 10"><path d="M9.79.212A.72.72 0 0 0 8.77.211L5 3.981 1.229.211A.717.717 0 0 0 .211.212H.21A.72.72 0 0 0 .211 1.23l3.771 3.769L.211 8.77a.718.718 0 0 0 0 1.018.718.718 0 0 0 1.018 0L5 6.017l3.771 3.771A.72.72 0 0 0 9.789 8.77L6.018 4.999 9.789 1.23A.72.72 0 0 0 9.79.212"></path></svg>
                     </span>
                 </div>
-                <div class="facet-header-title">Facet title</div>
+                <div class="facet-header-title text-medium-blue bold">Facet title</div>
                 <div class="facet-header-wait-animation" style="display: none;"></div>
             </div>
             <ul class="facet-values">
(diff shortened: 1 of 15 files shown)


docs/_components/components/10_facet.html
```diff
@@ -0,0 +1,159 @@
...
+usage: Facet styling
+stylesheet: '/scss/components/facet.scss'
+---
+<div class="flex">
+    <div class="facet-column">
+        <div class="facet">
+            <div class="facet-header">
+                <div class="facet-header-eraser hidden">
+                    <span class="icon fill-medium-grey">
+                        <svg viewBox="0 0 10 10"><path d="M9.79.212A.72.72 0 0 0 8.77.211L5 3.981 1.229.211A.717.717 0 0 0 .211.212H.21A.72.72 0 0 0 .211 1.23l3.771 3.769L.211 8.77a.718.718 0 0 0 0 1.018.718.718 0 0 0 1.018 0L5 6.017l3.771 3.771A.72.72 0 0 0 9.789 8.77L6.018 4.999 9.789 1.23A.72.72 0 0 0 9.79.212"></path></svg>
```
GermainBergeron: You can use `{{ site.data.icons['icon-name'] }}` instead of the svg to keep the template simple

scss/components/facet.scss
```diff
@@ -0,0 +1,183 @@
+.facet-column {
+  display: block;
+  width: 30%;
+  max-width: 360px;
+  padding: 24px 32px 24px 40px;
```
GermainBergeron: I think we should use variables for this

GermainBergeron:
Looks really good!

wfortin:
Awesome

scss/components/facet.scss
```diff
@@ -0,0 +1,183 @@
...
+      fill: $dark-grey;
+    }
+  }
+}
+
+.facet-header-title {
+  float: left;
+
+  font-family: $base-font-family;
+  font-size: 16px;
```
pochretien: i think we can create a global variable for the font size like 'title font size : 16' and default font size : 15, something like that to scale the facet with the vapor font size

scss/components/facet.scss
```diff
@@ -0,0 +1,183 @@
...
+  display: inline-block;
+  float: left;
+  width: 18px;
+  height: 18px;
+  margin-right: 15px;
+
+  text-align: center;
+  vertical-align: middle;
+
+  background: white;
```
pochretien: variable `@pure-white`

scss/components/facet.scss
```diff
@@ -0,0 +1,183 @@
...
+
+.facet-value-exclude {
+  position: absolute;
+  top: 9px;
+  right: 5px;
+
+  width: 13px;
+  height: 13px;
+
+  background: white;
```
pochretien: same here for variable

pochretien:
very nice :gg:

PR #201, opened 2016-07-26
Collapsible rows rework

PMarcL:


scss/tables/table-collapsible-rows.scss
```diff
@@ -42,6 +42,15 @@ table {
           .container {
             display: none;
           }
+
+          h4 {
+            padding-top: 10px;
+
+            font-size: 12px;
```
JPLachance: Why would we change the h4 default font size? Maybe we should not use h4.

scss/tables/table-collapsible-rows.scss
```diff
@@ -42,6 +42,15 @@ table {
           .container {
             display: none;
           }
+
+          h4 {
+            padding-top: 10px;
```
JPLachance: Add a variable.

scss/tables/table-collapsible-rows.scss
pochretien: +1 keep the default font-size for h4 or if u need less, use a other one

scss/tables/table-collapsible-rows.scss
```diff
@@ -42,6 +42,15 @@ table {
...
+
+            font-size: 12px;
+            font-weight: bold;
+            text-transform: uppercase;
```
pochretien: can u confirm with UX, in the mockup, the collapsable section title are in TitleCase and now we use Uppercase

GermainBergeron:
run `gulp sass:format` then :shipit:

PR #207, opened 2016-08-04
fixed filter box on small screens

PMarcL:


scss/controls/filter-box.scss
```diff
@@ -5,14 +5,14 @@
     width: 260px;
 
     .filter-icon {
-      right: 0;
+      right: 10px;
```
JPLachance: Add variables?

PMarcL: Yes, you're right. In fact, I'll add a bunch of variable in this file since we're using constants everywhere.

PR #208, opened 2016-08-05
Facet box with more facet values

sroy3:
+ Added a 'more' checkbox to toggle more facet row values with a filter box

- Removed unused exclude button

screen shot 2016-08-05 at 10 04 36 am

### DIFF
--- a/docs/_components/components/10_facet.html
+++ b/docs/_components/components/10_facet.html
@@ -30,9 +30,6 @@ stylesheet: '/scss/components/facet.scss'
                             <span class="facet-value-caption">First facet value</span>
                         </div>
                     </label>
-                    <div class="facet-value-exclude">
-                        <span class="icon"></span>
-                    </div>
                 </li>
                 <li class="facet-value facet-selectable">
                     <label class="facet-value-label">
@@ -47,9 +44,6 @@ stylesheet: '/scss/components/facet.scss'
                             <span class="facet-value-caption">Second facet value</span>
                         </div>
                     </label>
-                    <div class="facet-value-exclude">
-                        <span class="icon"></span>
-                    </div>
                 </li>
                 <li class="facet-value facet-selectable">
                     <label class="facet-value-label">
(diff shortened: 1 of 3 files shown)


wfortin:
:shipit:

docs/_components/components/10_facet.html
```diff
@@ -64,9 +58,197 @@
...
+                                <span></span>
+                            </div>
+                            <span>More</span>
+                        </div>
+                    </label>
+                </li>
+            </ul>
+            <div class="facet-more-search hidden">
+                <div class="facet-search">
+                    <div class="coveo-filter-container">
```
JPLachance: Instead of doing a custom dropdown, will it be possible to use our chosen single select? It's just a question.

scss/components/facet.scss
```diff
@@ -119,6 +129,39 @@
...
+
+    content: '';
+
+    background-color: $medium-grey;
+    transform: translate(-50%, -50%);
+  }
+
+  &:before {
+    width: 2px;
+    height: 70%;
```
JPLachance: Why 70%? Plz use variables for the width and the height in both the before and the after.

scss/components/facet.scss
```diff
@@ -166,18 +209,41 @@
...
+  }
+}
+
+.facet-search-results {
+  position: absolute;
+  top: calc(100% - 25px);
+  left: 20px;
+  z-index: 999;
+
+  width: calc(100% - 40px);
```
JPLachance: - 40px ? I gess it's for the left/right padding. As said, it should be 16px, not 20. The mockup was wrong. Plz update for the $button-padding-x variable. ($button-padding-x * 2)

scss/components/facet.scss
sroy3: Is it better to keep the size ratio of the checkbox (70%) and plus icon or to keep the padding around the plus icon (calc(100% - 6px)) if we ever decide to change the size of the checkbox? Or we can simply decide on using 12px. I decided on using 70% its ratio, but the other 2 options are no worse in my opinion.

PR #210, opened 2016-08-09
fix margin for radio button and checkbox

pochretien:


scss/controls/checkboxes.scss
```diff
@@ -134,7 +134,7 @@ label[for].coveo-checkbox-label {
     clear: left;
 
     & + .coveo-checkbox-label {
-      margin-top: 18px;
+      margin-top: 10px;
```
PMarcL: Use a variable maybe?

wfortin: +1

PR #219, opened 2016-08-23
made the inputs lighter when disabled

PMarcL:


scss/forms/table-form.scss
```diff
@@ -59,12 +59,13 @@
       background-color: $white;
 
       input:disabled {
-        color: black;
```
PMarcL: I don't know why we made the input darker when disabled. I investigated to see if that was actually used in the admin. Didn't find anything.

PMarcL: It's use when we try to edit a source after creation, the input for the name of the source is disabled.  It still makes more sens the make the text lighter with the opacity when it's disabled then to make it darker IMO.

PR #220, opened 2016-08-26
+ Added a vertically aligned copy of the scale out animation

sroy3:
For elements that have `translateY(-50%)` since `transform: scale(0.0)` overwrites the transform property.


pochretien:
🚢 it

PR #221, opened 2016-08-29
coveo-child update!

JPLachance:
* Update the coveo-child arrow and background color
* Update the $white palette color for the new one
* Make the radio button label display flex to simplify usage
screen shot 2016-08-29 at 12 02 45 pm


wfortin:
:shipit:

PR #222, opened 2016-08-29
+ Added a file input form control

sroy3:
screen shot 2016-08-29 at 4 19 57 pm

screen shot 2016-08-29 at 4 20 20 pm

### DIFF
--- a/scss/controls/file-input.scss
+++ b/scss/controls/file-input.scss
@@ -6,16 +6,6 @@
 
   .file-path-wrapper {
     flex-grow: 1;
-
-    input {
-      border-bottom-color: $medium-grey;
-
-      @include placeholder();
-
-      &.has-file {
-        border-bottom: 1px solid $medium-blue;
-      }
-    }
   }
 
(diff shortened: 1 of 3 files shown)


scss/controls/file-input.scss
```diff
@@ -0,0 +1,37 @@
...
+    input {
+      position: absolute;
+      top: 0;
+      left: 0;
+
+      width: 100%;
+      height: 100%;
+
+      cursor: pointer;
+      opacity: 0;
```
wfortin: Is position absolute really needed since we are in a flex container ?
Cant we juste use `align-self: flex-start` ?

sroy3: It's the hidden file input so that wherever we click, we open the choose file window.

wfortin: Ha makes more sense.

wfortin:
🚢 it

pochretien:
🚢 🇮🇹

PR #227, opened 2016-09-06
+ Added a max-width on table cells

sroy3:
So that long non-breaking strings get cut and show ellipsis instead

screen shot 2016-09-06 at 2 26 03 pm

### DIFF
--- a/scss/tables/table.scss
+++ b/scss/tables/table.scss
@@ -147,6 +147,13 @@ table:not(.datepicker-table) {
 
     border-bottom: $table-border;
 
+    * {
+      max-width: 400px;
+      overflow: hidden;
+
+      text-overflow: ellipsis;
+    }
+
     &:first-child {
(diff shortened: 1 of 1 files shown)


scss/tables/table.scss
```diff
@@ -147,6 +147,13 @@ table:not(.datepicker-table) {
 
     border-bottom: $table-border;
 
+    * {
```
JPLachance: But, I'm really not sure we want to add that limitation to EVERY table cell... I can easily see a use-case where we have a table with two columns, one with only an icon in it and the other that take the rest of the table width.. Therefore, I think it would be a better idea to add a modifier, `mod-max-width`, instead of forcing every table cells to have that max width.

wfortin: +1 to add that to a modifier and add that to columns that are user input or knowingly long.

sroy3: You are right. Done.

PR #233, opened 2016-09-19
Popovers and small buttons!

JPLachance:
+ Add the mod-small for buttons
+ Add the popover with his documentation
* Replace some components in the documentation
screen shot 2016-09-19 at 3 47 12 pm
screen shot 2016-09-19 at 4 08 26 pm


scss/components/popover.scss
```diff
@@ -0,0 +1,24 @@
+.popover {
+  min-width: $popover-min-width;
+
+  background-color: $pure-white;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.4);
```
pochretien: u can use rgba($pure-white, 0.4) i think

scss/components/popover.scss
```diff
@@ -0,0 +1,24 @@
...
+  border: 1px solid $greyish-dark-blue;
+  border-radius: 2px;
```
pochretien: use the variable

pochretien:
:gg:

scss/components/popover.scss
GermainBergeron: It's black, not white ;)

JPLachance: Done!

scss/components/popover.scss
JPLachance: Done!

PR #240, opened 2016-10-04
Simple spaced boxes!

JPLachance:
Add a simple spaced boxes layout utility and a simple demo!


scss/variables.scss
```diff
@@ -170,3 +170,6 @@ $facet-column-padding-left: 40px;
 $facet-checkbox-mark-size: 12px;
 $facet-checkbox-line-width: 2px;
 $facet-more-height: 200px;
+
+// Spaced boxes
+$box-spacing: 10px;
```
pochretien: I think we can use the same name $space-box-spacing

pochretien: I think we can change //Spaced boxes and create a section to set all generic variable like this spacing.

PR #260, opened 2016-10-20
Big table

PMarcL:


scss/tables/big-table.scss
```diff
@@ -0,0 +1,77 @@
...
+  tbody {
+    tr {
+      height: $big-table-height;
+
+      cursor: default;
+
+      td {
+        padding: $table-cell-left-right-padding;
+
+        vertical-align: top;
```
JPLachance: Is this really necessary?

scss/tables/big-table.scss
```diff
@@ -0,0 +1,77 @@
...
+
+      td.main {
+        white-space: nowrap;
+
+        .icon {
+          width: $big-table-height;
+          height: $big-table-height;
+          margin-right: 20px;
+
+          vertical-align: -3px;
```
JPLachance: Can we do something better here ?

scss/tables/big-table.scss
```diff
@@ -0,0 +1,77 @@
...
+        vertical-align: top;
+      }
+
+      td.main {
+        white-space: nowrap;
+
+        .icon {
+          width: $big-table-height;
+          height: $big-table-height;
+          margin-right: 20px;
```
JPLachance: I would use a variable here.

scss/tables/big-table.scss
```diff
@@ -0,0 +1,77 @@
...
+
+          vertical-align: -3px;
+        }
+
+        div.details {
+          display: inline-block;
+
+          div.second-row {
+            .icon {
+              width: 15px;
```
JPLachance: Missing variables for the icon size.

scss/tables/big-table.scss
```diff
@@ -0,0 +1,77 @@
...
+            }
+          }
+        }
+      }
+
+      td.col, td.main {
+        position: relative;
+
+        .first-row {
+          font-size: 15px;
```
JPLachance: You should use a variable here for the font-size and the line height.

scss/tables/big-table.scss
```diff
@@ -0,0 +1,77 @@
...
+          line-height: 20px;
+        }
+
+        div.second-row {
+          margin-top: 2px;
+
+          font-size: 12px;
```
JPLachance: Same here, use variables.

scss/tables/big-table.scss
```diff
@@ -0,0 +1,77 @@
...
+      td.col, td.main {
+        position: relative;
+
+        .first-row {
+          font-size: 15px;
+          line-height: 20px;
+        }
+
+        div.second-row {
+          margin-top: 2px;
```
JPLachance: Should add a comment (and or variable)

scss/tables/big-table.scss
```diff
@@ -0,0 +1,77 @@
...
+
+          font-size: 12px;
+          line-height: 17px;
+        }
+      }
+
+      div.wrapper {
+        min-height: $big-table-height;
```
JPLachance: big-table-height ? big-table-row-height would sound better to me.

scss/tables/big-table.scss
```diff
@@ -0,0 +1,77 @@
...
+      }
+
+      &.disabled td .wrapper {
+        opacity: 0.5;
+      }
+    }
+
+    tr.error-blankslate-row {
```
JPLachance: I don't think this should be in the `big-table.scss` file, but that's only my opinion. Your choice.

scss/tables/big-table.scss
```diff
@@ -0,0 +1,77 @@
...
+      &:hover {
+        background-color: $pure-white;
+      }
+
+      td.error-blankslate-section {
+        padding: $big-table-blankslate-section-padding;
```
JPLachance: In my opinion, a blank slate in a normal table or a big table is the same thing. I think we are too specific here.

JPLachance:
And, I would love a "big table" example in the Jekyll doc :)

PMarcL:
Yeah, I thought about it, but setting up ruby and jekyll on pc is a nightmare and it didn't work on my work laptop... I'll try with my personnal laptop.

wfortin:
Yeah Jekyll is a pain on windows. Might want to check https://github.com/coveo/source.coveo.com#run-locally for a docker version of jekyll.

PR #268, opened 2016-11-03
* Modified the mod-max-width to be smaller

sroy3:


scss/tables/table.scss
```diff
@@ -148,7 +148,7 @@ table:not(.datepicker-table) {
     border-bottom: $table-border;
 
     .mod-max-width {
```
JPLachance: Question, what is this used for?

PR #270, opened 2016-11-04
+ Added loading prompt and bouncing loading

sroy3:
Moved from the admin to be used in react-vapor.


wfortin:
Nice! I'd like to see a corresponding react-vapor component.

PR #273, opened 2016-11-11
Moved pagination from admin to be used in react vapor

sroy3:


scss/guide.scss
```diff
@@ -45,6 +45,7 @@
   @import 'tables/table-actions';
   @import 'tables/table-align-header';
   @import 'tables/table-collapsible-rows';
+  @import 'tables/table-navigation';
```
GermainBergeron: I think we should name the file table-pagination or simply pagination (it's not always used in a table)

docs/_components/tables/05_table-navigation.html
```diff
@@ -0,0 +1,100 @@
+---
+type: tables
+title: Table navigation
+usage: Adds buttons to change the number of items per page and change page.
+stylesheet: '/scss/tables/table-collapsible-rows.scss'
```
GermainBergeron: wrong stylesheet

scss/guide.scss
sroy3: Done

docs/_components/tables/05_table-navigation.html
sroy3: Oops, changed.

PR #275, opened 2016-11-18
* Changed some specs for the modal promp, checkboxes and radio buttons

sroy3:


scss/components/collapsible.scss
```diff
@@ -20,7 +21,7 @@
 
     cursor: pointer;
 
-    &:focus, &:hover {
```
GermainBergeron: No more style on hover?

sroy3: UX team decided to remove it, but you made me doubt and read again. It also needs to be blue when expanded even if not focused. I'll need to change that.

sroy3: Done

scss/variables.scss
```diff
@@ -21,11 +21,11 @@ $button-confirm-icon-width: 15px;
 $button-confirm-margin-bottom: 20px;
 
 // Checkboxes and radios
-$checkbox-size: 16px;
-$checkbox-label-margin: 16px;
+$checkbox-size: 18px;
```
JPLachance: Is this breaking stuff in the UI? Like group privileges table? Or the Salesforce form?

sroy3: I re-aligned them.

PR #276, opened 2016-11-28
* Updated facet style

sroy3:
screen shot 2016-11-28 at 11 35 40 am

### DIFF
--- a/scss/controls/checkboxes.scss
+++ b/scss/controls/checkboxes.scss
@@ -5,7 +5,7 @@
   height: $checkbox-size;
   padding: 0;
 
-  background: $pure-white none 0 0;
+  background: $pure-white;
   border: 2px solid $medium-blue;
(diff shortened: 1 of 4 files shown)


scss/controls/checkboxes.scss
```diff
@@ -1,21 +1,25 @@
...
+.coveo-checkbox-button {
+  position: relative;
 
-  & + button {
-    position: relative;
+  width: $checkbox-size;
+  height: $checkbox-size;
+  padding: 0;
+
+  background: $pure-white none 0 0;
```
GermainBergeron: I think you can simply use `background: $pure-white;`

sroy3: Hmm, I haven't changed this section, I simply moved stuff to be re-used. I will investigate why it was this way and change it if it has no use (most likely).

PR #282, opened 2016-12-14
add mod-description and docs for small button

pochretien:
add a new mod for the p tag
add a new button size for the docs

### DIFF
--- a/docs/_components/form_controls/01_input.html
+++ b/docs/_components/form_controls/01_input.html
@@ -75,6 +75,13 @@ stylesheet: '/scss/controls/input-field.scss'
         </label>
         <span class="generic-form-error">Error message goes here!</span>
     </div>
+
+    <div id="description" class="input-field form-group">
+        <h4 class="text-medium-blue">Simple description typography</h4>
+        <p class="mod-description">Lorem ipsum dolor sit amet, interdum est eu. A tellus condimentum hendrerit enim, ligula fusce vitae, leo
+            et fusce mauris lorem suscipit scelerisque. Vel arcu, non vestibulum suspendisse non lectus magnis suspendisse, aliquam magna
+            commodo. Nascetur eleifend at faucibus faucibus, nulla fringilla, mauris ultrices posuere in, ut dolor cras rhoncus.</p>
+    </div>
 </div>
 
 <script>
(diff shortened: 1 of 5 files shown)


docs/_components/form_controls/01_input.html
```diff
@@ -75,6 +75,13 @@
         </label>
         <span class="generic-form-error">Error message goes here!</span>
     </div>
+
+    <div id="description" class="input-field form-group">
+        <h4 class="text-medium-blue">Simple description typography</h4>
+        <p class="mod-description">Lorem ipsum dolor sit amet, interdum est eu. A tellus condimentum hendrerit enim, ligula fusce vitae, leo
```
JPLachance: Why adding this while it should only be our basic paragraph style? I would simply create a class named `paragraph` (or something else if anyone has a better idea).

JPLachance: I'm not saying you should set the style right on the `p` tag, I'm saying you should have a class `.paragraph`, just like we have a class `.btn`. `mod-` class is for when you're modifying another existing class.

docs/_components/form_controls/01_input.html
```diff
@@ -75,6 +75,13 @@
         </label>
         <span class="generic-form-error">Error message goes here!</span>
     </div>
+
+    <div id="description" class="input-field form-group">
```
JPLachance: Why add an id on this div?

scss/typography/elements.scss
```diff
@@ -0,0 +1,5 @@
+p.mod-description {
```
JPLachance: Here I would only do 
```
.paragraph {
  color: $dark-grey;
   font-size: $default-font-size;
   line-height: $big-line-height;
 }
```

scss/typography/elements.scss
```diff
@@ -0,0 +1,5 @@
+p.paragraph {
```
vseguin: I'm wondering if we want to scope this on the p element... what are we usually doing?

docs/_components/form_controls/01_input.html
vseguin: Same question.

scss/typography/elements.scss
JPLachance: @sroy3 what do you think?

sroy3: I'd remove the p. It is useless and makes for harder future changes.

vseguin: OMG I WAS RIGHT ON A STYLE COMMENT, IS THIS CHRISTMAS

PR #284, opened 2016-12-22
Fix alignment with btn-container and btn tag

pochretien:
add new page to show the button alignment
add css for btn-container and btn alignment
add exemple for the tooltip on a tag a and btn disabled


pochretien:
screen shot 2016-12-21 at 8 14 54 pm

PR #285, opened 2016-12-22
Panel header fix

PMarcL:


scss/variables.scss
```diff
@@ -64,6 +64,8 @@ $header-padding: 40px;
 $header-height-padding: 20px;
 $header-padding-y: 5px;
 $description-margin-top: 5px;
+$panel-header-height: 100px;
+$description-min-height: 38px;
```
vseguin: You could at least put descriptions variables together i'd say

PR #289, opened 2017-01-13
Fix popover height!

JPLachance:
! Fix the height of the modal-content class (there was a bug when showing a backdrop over a popover)

Fixes this bug (the double backdrop):
screen shot 2017-01-13 at 3 04 48 pm

### DIFF
--- a/scss/components/prompt.scss
+++ b/scss/components/prompt.scss
@@ -26,7 +26,7 @@
         flex-basis: initial;
         flex-grow: initial;
 
-        min-height: calc(#{$modal-prompt-min-height} - #{$modal-prompt-header-height} - #{$modal-prompt-footer-height}); // Use all the space left
+        min-height: $modal-prompt-body-min-height;
 
         &.mod-header-padding {
           padding-right: $modal-prompt-horizontal-padding;
--- a/scss/components/prompt.scss
+++ b/scss/components/prompt.scss
@@ -1,6 +1,7 @@
 .modal-container {
   &.mod-prompt {
     .modal-content {
+      height: auto; // Required or modal-content height is always equal to $modal-prompt-max-height.
       max-width: $modal-prompt-max-width;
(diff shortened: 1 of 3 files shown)


scss/variables.scss
```diff
@@ -82,11 +82,12 @@ $modal-prompt-body-max-height: 120px;
 $modal-prompt-body-min-height: 80px;
```
sroy3: Is this still used? You could move the min height calculation (`calc(#{$modal-prompt-min-height} - #{$modal-prompt-header-height} - #{$modal-prompt-footer-height})`) if it is not.

scss/variables.scss
```diff
@@ -78,15 +78,14 @@ $modal-footer-padding: 24px 40px;
 
 // Prompt modal
 $modal-prompt-text-padding-y: 24px;
-$modal-prompt-body-max-height: 120px;
-$modal-prompt-body-min-height: 80px;
```
sroy3: You could still use `$modal-prompt-body-min-height` and move the min height calculation (`calc(#{$modal-prompt-min-height} - #{$modal-prompt-header-height} - #{$modal-prompt-footer-height})`) here.

scss/components/prompt.scss
```diff
@@ -21,8 +22,11 @@
       }
 
       .modal-body {
-        max-height: $modal-prompt-body-max-height;
-        min-height: $modal-prompt-body-min-height;
+        // Reset flex-basis and flex-grow values so autosize based on content can work.
+        flex-basis: initial;
+        flex-grow: initial;
+
+        min-height: calc(#{$modal-prompt-min-height} - #{$modal-prompt-header-height} - #{$modal-prompt-footer-height}); // Use all the space left
```
pochretien: u can remove this comment IMO

scss/variables.scss
JPLachance: True.

Issue #136, opened 2016-04-18
Generic name space everywhere or nowhere


apare:
I'm not sure if we should remove all namespace or add them everywhere

For example we have `coveo-checkbox` and `coveo-form` but we also have `input-field` `form-group`

I think we should go with Coveo everywhere because this reduce the chance to match a external lib css (Salesforce)

JPLachance:
After a discussion with @wfortin, we will remove the `coveo-` prefix everywhere and bump a major version.

PR #290, opened 2017-01-16
Item filter

sroy3:
screen shot 2017-01-16 at 11 03 53 am

### DIFF
--- a/scss/tables/table-actions.scss
+++ b/scss/tables/table-actions.scss
@@ -100,18 +100,18 @@
   }
 }
 
-.coveo-item-filter {
+.item-filter {
   flex-grow: 1;
 }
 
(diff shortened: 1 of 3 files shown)


scss/tables/table-actions.scss
```diff
@@ -99,3 +99,24 @@
     padding-right: $header-padding;
   }
 }
+
+.coveo-item-filter {
```
JPLachance: Eeee question: are we putting `coveo-` everywhere or not? See https://github.com/coveo/vapor/issues/136

sroy3: I had forgotten about that. I will remove it. Thanks. I think it is still a good idea to remove it everywhere and release a major version.

PR #292, opened 2017-01-17
- Removed the admin description min height

sroy3:


scss/components/panel-header.scss
```diff
@@ -11,7 +11,6 @@
   }
 
   .admin-description {
-    min-height: $description-min-height;
```
GermainBergeron: is the variable still used?

PR #297, opened 2017-01-24
Feature modalheaderupdate

pochretien:
image


docs/_components/components/08_modal.html
```diff
@@ -17,8 +18,13 @@
         <header class="modal-header">
             <h1>Basic modal</h1>
             <span class="small-close js-modal-close">
-                <svg width="512" height="512" viewBox="0 0 512 512" class="icon fill-pure-white">
-                    <path d="M480.002 96l-64-63.998-160 159.998-160-160-64 63.998L192.004 256l-160 159.998 64 64 160-159.998 159.994 159.994 64-64L320.003 256"></path>
+                <svg viewBox="0 0 22 22" class="icon fill-pure-white">
```
JPLachance: Why did this changed? For what I can see, the icon code is not clean. The `<g transform="translate(1.000000, 1.000000)">` is useless (doing nothing) and the other `<g>` with no properties on it is also useless. I would keep the previous icon or pass this one through SVGCleaner.

docs/_components/components/08_modal.html
```diff
@@ -35,8 +41,13 @@
         <header class="modal-header">
             <h1>Mod small</h1>
             <span class="small-close js-modal-close">
-                <svg width="512" height="512" viewBox="0 0 512 512" class="icon fill-pure-white">
-                    <path d="M480.002 96l-64-63.998-160 159.998-160-160-64 63.998L192.004 256l-160 159.998 64 64 160-159.998 159.994 159.994 64-64L320.003 256"></path>
+                <svg viewBox="0 0 22 22" class="icon fill-pure-white">
```
JPLachance: Same comment here and for the rest of the file. Keep the previous icon or use SVG cleaner. If you change the icon, **make sure it is working in all browsers**.

scss/components/modal.scss
```diff
@@ -1,5 +1,6 @@
 $base-backdrop-z-index: 1049;
 $base-modal-z-index: 1050;
+$icon-size: (4em / 3);
```
JPLachance: _Why use em here instead of px?_ => Ho, okay, it was like that before.. Louche! Maybe we should change that?

pochretien: // Make both the icon and his containing span the same size.  ....

docs/_components/components/08_modal.html
pochretien: Done! new svg is now cleaner

JPLachance: The `transform="matrix(.7071-.7071.7071.7071-3.142 11)` is still weird IMO. What's the diffenrence between the old icon and the new one?

scss/components/modal.scss
GermainBergeron: If you use a size in em, it's relative to the container font size. If you use pixels it will look weird on bigger/smaller fonts

PR #302, opened 2017-01-27
add the pure-white color for the prompt header text

pochretien:


publish-pages.sh
```diff
@@ -4,8 +4,8 @@ set -x
 
 ( cd _gh_pages
   git init
-  git config user.name "pastjean"
-  git config user.email "pastjean@coveo.com"
+  git config user.name "pochretien"
+  git config user.email "pochretien@coveo.com"
```
sroy3: Why do you have to change the user for it to work?

pochretien: I simply added someone from the team working on it

GermainBergeron: We should probably ask devbooster for a user/user email for this

MichelML: @sroy3 personally it was not working when I tried it last time with pastjean

PR #303, opened 2017-01-27
add value for ie11 to avoid the double calc

pochretien:
fix the body size in ie11

### DIFF
--- a/scss/components/prompt.scss
+++ b/scss/components/prompt.scss
@@ -26,9 +26,10 @@ $prompt-message-line-height: 20px;
 
       .modal-body {
         // Reset flex-basis and flex-grow values so autosize based on content can work.
-        flex-basis: initial;
+        flex-basis: inherit;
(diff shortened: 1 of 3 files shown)


pochretien:
image

image

scss/components/prompt.scss
```diff
@@ -26,9 +26,10 @@ $prompt-message-line-height: 20px;
 
       .modal-body {
         // Reset flex-basis and flex-grow values so autosize based on content can work.
-        flex-basis: initial;
+        flex-basis: inherit;
```
JPLachance: Hum, what does that change?

pochretien: For ie11, the modal-body can now take the size of his parent. This change avoid the content to always take the min-height. Nothing changes for other browsers.

JPLachance: okay!

PR #305, opened 2017-02-06
Date picker

sroy3:
screen shot 2017-02-06 at 3 53 06 pm

### DIFF
--- a/resources/icons/svg/set-to-now.svg
+++ b/resources/icons/svg/set-to-now.svg
@@ -1,1 +1,1 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 18 18" enable-background="new 0 0 18 18"><g transform="translate(-7-8)"><path d="m7 9h18v16h-18v-16m2 4v10h14v-10h-14m9 6h3v2h-3v-2"/></g></svg>
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 18 18"><g transform="translate(-7-8)"><path d="m7 9h18v16h-18v-16m2 4v10h14v-10h-14m9 6h3v2h-3v-2"/></g></svg>
--- a/resources/icons/svg/set-to-now.svg
+++ b/resources/icons/svg/set-to-now.svg
@@ -1,0 +1,1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 18 18" enable-background="new 0 0 18 18"><g transform="translate(-7-8)"><path d="m7 9h18v16h-18v-16m2 4v10h14v-10h-14m9 6h3v2h-3v-2"/></g></svg>
--- a/scss/components/calendar.scss
+++ b/scss/components/calendar.scss
@@ -260,6 +260,10 @@
       color: $pure-white;
     }
   }
+
+  > span {
+    order: 1;
+  }
 }
 
 .mod-inline {
--- a/scss/components/calendar.scss
+++ b/scss/components/calendar.scss
@@ -46,11 +46,11 @@
       color: $dark-grey;
       text-transform: uppercase;
 
-      border-bottom: 1px solid $light-grey;
+      border-bottom: $calendar-border;
     }
 
     td {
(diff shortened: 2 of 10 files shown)


scss/components/calendar.scss
```diff
@@ -0,0 +1,305 @@
...
+
+    th {
+      color: $dark-grey;
+      text-transform: uppercase;
+
+      border-bottom: 1px solid $light-grey;
+    }
+
+    td {
+      border: 1px solid $light-grey;
```
JPLachance: Don't we have that (`1px solid $light-grey`) inside a variable somewhere?

scss/components/calendar.scss
```diff
@@ -0,0 +1,305 @@
...
+
+  &.mod-inline {
+    display: inline-block;
+  }
+}
+
+.options-cycle-button {
+  padding: 0;
+
+  font-size: 0.8em;
```
JPLachance: Why use em here?

scss/components/calendar.scss
JPLachance: $default-border IIRC.

scss/components/modal.scss
```diff
@@ -209,3 +136,80 @@ $icon-size: (4em / 3);
     transform: translate3d(0, 0, 0);
   }
 }
+
+.modal-header {
```
JPLachance: Un-scoping those may break things at a couple of places (where we had legacy style). For example, if someone tries to copy that file and use it in V1, it will break for sure. Just do a quick tour of our projects before merging that.

scss/guide.scss
```diff
@@ -8,6 +8,16 @@
 @import 'variables';
 @import 'animations';
 @import 'helpers';
+// Utility
+@import 'typography/utility';
```
JPLachance: @tedre191 used Vapor inside Salesforce and was complaining about some of our un-scoped css. Will this affect the Salesforce integration?

sroy3: All utilities depend only on the files in the previous block. For some reason they were placed at the end and could not be used in other files. We can now extend on base classes used in those files or add mixins and functions like I did.

scss/components/calendar.scss
sroy3: Because the option cycle can be of different sizes, I make sure the buttons have the same size ratio with the title.

scss/components/calendar.scss
sroy3: $default-border is $medium-grey, but I added a new variable for this one.

scss/components/modal.scss
sroy3: It simply means the style will apply to modal-header, modal-footer... even if they aren't in a modal-container. These aren't very generic classes, they weren't used for anything else but for what they are described to do here. If anything, it simply makes it easier to overwrite these styles. Also, this is not used in V1.

resources/icons/svg/set-to-now.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 18 18" enable-background="new 0 0 18 18"><g transform="translate(-7-8)"><path d="m7 9h18v16h-18v-16m2 4v10h14v-10h-14m9 6h3v2h-3v-2"/></g></svg>
```
pochretien: can remove enable-background

scss/utility/colors.scss
```diff
@@ -15,4 +15,8 @@ $palette-secondary: red $red, yellow $yellow, green $green, blue $blue, salesfor
   .text-#{nth($color, 1)} {
     color: nth($color, 2);
   }
+
+  .border-#{nth($color, 1)} {
```
pochretien: nice!

pochretien:
Juste make sure the date pick look good in ie 11 and ship it!

resources/icons/svg/set-to-now.svg
sroy3: Removed

PR #307, opened 2017-02-07
Fix the coveo-checkbox-labels-group to allow coveo-child to be properly used

JPLachance:
- Remove the coveo-checkbox-labels-group `float: left;` hack (replace by display flex (instead of inline flex when outside of a group)

+ Add the coveo-checkbox-group-option class for the checkbox group spacing handling to support having a cover-parent inside of a checkbox group

! Also fix the mod-fixed-width for the coveo-child


scss/controls/checkboxes.scss
```diff
@@ -136,14 +136,16 @@ label[for].coveo-checkbox-label {
...
 .coveo-checkbox-labels-group {
-  @extend .clearfix;
-
   .coveo-checkbox-label {
-    float: left;
-    clear: left;
+    display: flex;
 
+    // Deprecated, use the coveo-checkbox-group-option class instead.
     & + .coveo-checkbox-label {
```
MichelML: Do you leave it there because it might break things if you remove it entirely?

JPLachance: Yup, removing it would require to update every project using it. We always prefer to avoid breaking changes and one day, we will do a major release and remove every deprecated pieces of code.

PR #308, opened 2017-02-09
Add new icons for groups and user

GermainBergeron:


resources/icons/svg/identity-group.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 12">
+<g fill="currentColor">
```
pochretien: u can remove this tag with the fill

PR #314, opened 2017-02-22
Make sure the error row is the right color in the alternating rows

sroy3:


scss/tables/table-alternating-rows.scss
```diff
@@ -16,6 +16,10 @@ table.mod-alternate {
       tr, tr:hover, tr.collapsible-row, tr.collapsible-row:hover {
         td, td:hover {
           background-color: $white;
+
+          .source-activity-error-container {
```
GermainBergeron: Should we do the fix in the admin instead? It seems a bit specific for vapor

sroy3: I should have noticed this. Although I think it would be best to move the style of the error to vapor since it is being used many places and react vapor is one of them. I'll do that (and change the class names as well).

PR #318, opened 2017-03-14
Added Sync Feedback

GermainBergeron:
Added style & examples

sync-feedback


sroy3:
The saving changes flashing dot seems to be a little more to the right than the check or the x. Are they the same size?

GermainBergeron:
The SVG size is the same as icon size (13x13px), I think the SVGs might have a small gap or it might be an illusion. Since we don't usually see the different status next to each other I wouldn't change it

PR #322, opened 2017-03-17
Migrates tabs from jsadmin-v2

sbelzile-Coveo:


docs/_components/components/17_tab.html
```diff
@@ -0,0 +1,22 @@
...
+        <div class="mod-header-padding mod-form-top-bottom-padding">
+            Content of Tab 1
+        </div>
+    </div>
+    <div class="tab-pane">
+        <div class="mod-header-padding mod-form-top-bottom-padding">
+            Content of Tab 2
+        </div>
+    </div>
+</div>
```
GermainBergeron: You can add a small script under the div to change the active tab/pane, it's nice when we can interact with the components

scss/components/tab.scss
```diff
@@ -0,0 +1,73 @@
...
+
+    visibility: visible;
+
+    @extend .slideInLeft;
+  }
+
+  .loading, & > .spinner {
+    margin-top: 50px;
+  }
+}
```
GermainBergeron: I guess this is taken directly from v2 without changes?

sbelzile-Coveo: exact

scss/animations.scss
```diff
@@ -46,3 +46,24 @@
...
+    -webkit-transform: translateX(0);
+    -ms-transform: translateX(0);
+    transform: translateX(0);
+  }
+}
+
+.slideInLeft {
+  -webkit-animation-name: slideInLeft;
+  animation-name: slideInLeft;
+}
```
wfortin: Remove the -webkit, -ms, prefixes. Autoprefixer takes care of that.

scss/components/tab.scss
```diff
@@ -0,0 +1,73 @@
+.tab-navigation {
+  min-height: 36px;
+  padding: 0 $header-padding;
+
+  border-bottom: 1px solid $medium-grey;
```
pochretien: you can use $default-border

scss/components/tab.scss
```diff
@@ -0,0 +1,73 @@
+.tab-navigation {
+  min-height: 36px;
```
JPLachance: We prefer defining and using variables for all important properties like the height, the font size, margins, etc. See the variables.scss file.

PR #323, opened 2017-03-17
add validation for svg string

MichelML:


MichelML:
Right now there can be some cases where we pass a string to the param svgString when it is not a valid svgString (happened in v1), and can cause the UI to crash. This should be handled before the UI crashes. If it is not a valid SVG string, we thus return a default svg string when using the function svg() of VaporSVG.

resources/js/VaporSVG.js
```diff
@@ -49,6 +49,10 @@
         return s.camelize(s.humanize(svgName), true);
     }
 
+    function isSvgStringValid(svgString) {
+        return /^(<svg(\s.+)?>)(.+)?(<\/svg>)$/i.test(svgString);
```
GermainBergeron: You can simplify the regex a bit: `/^<svg(\s.+)?>.*<\/svg>$/` but it's really minor

PR #332, opened 2017-04-06
Fixed typos in css classes and updated dependencies

sroy3:


scss/components/prompt.scss
```diff
@@ -37,11 +37,16 @@ $prompt-message-line-height: 20px;
...
         }
 
-        .pompt-message {
+        .prompt-message {
           padding-top: $modal-prompt-text-padding-y;
 
           line-height: $prompt-message-line-height;
         }
+
+        .pompt-message {
```
vseguin: Is there any convention for deprecations? Or is it the first time we deprecate something in the styleguide?

sroy3: I search through the project (I hadn't think of it before) and it seems to have been done a couple of times before the exact same way that I did it.

vseguin: Excellent, just making sure that we stay consistent.

PR #334, opened 2017-04-07
Add icon 'clear-small

rousselm4:


resources/icons/svg/clear-small.svg
```diff
@@ -0,0 +1,5 @@
+<svg enable-background="new 0 0 12 12" viewBox="0 0 12 12" xmlns="http://www.w3.org/2000/svg">
+<ellipse cx="6" cy="6" fill="#fff" rx="5.377" ry="5.377" transform="matrix(.7071 -.7071 .7071 .7071 -2.4853 6)"/>
```
sroy3: You should remove the fill color here

rousselm4: It's only for background purpose. It's always supposed to be white. It's like this icon: https://github.com/coveo/vapor/blob/master/resources/icons/svg/help.svg

PR #338, opened 2017-04-25
Make sure panel headers with the class no-border-bottom do not add a border-bottom

sroy3:


scss/components/panel-header.scss
```diff
@@ -2,7 +2,9 @@
   min-height: $panel-header-height;
   padding: $header-padding-y $header-padding;
 
-  border-bottom: solid 1px $medium-grey;
+  &:not(.mod-no-border-bottom) {
+    border-bottom: solid 1px $medium-grey;
```
pochretien: u can use $default-border

PR #340, opened 2017-05-01
Increased the datepicker’s z-index

sroy3:
We currently cannot use the datepicker in a modal because of its low z-index.

### DIFF
--- a/scss/lib/datepicker.scss
+++ b/scss/lib/datepicker.scss
@@ -2,7 +2,7 @@ div.datepicker {
   position: absolute;
   top: 0;
   left: 0;
-  z-index: 5;
+  z-index: 5555;
 
   box-sizing: content-box;
   display: none;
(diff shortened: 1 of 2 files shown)


scss/lib/datepicker.scss
```diff
@@ -2,7 +2,7 @@ div.datepicker {
   position: absolute;
   top: 0;
   left: 0;
-  z-index: 5;
+  z-index: 5555;
```
vseguin: Any specific reason for such a specific number? :P

sroy3: Modals are in the thousands (1000, 1001...). I just repeated the number to get something higher. I know it's highly specific, but picking the right value for a z-index when other z-index are being chosen dynamically is quite hard.

pochretien: The z-index for the modal is 1050 and we increase by 20 for each layer added. Until the z-index is greater than 2000. We are safe

PR #342, opened 2017-05-03
add utils for the space-between

pochretien:
format sass


scss/elements/btn.scss
```diff
@@ -48,7 +48,7 @@
 
   cursor: pointer;
 
-  .icon:not([class*="fill-"]) {
```
pochretien: simple sass:format

PR #348, opened 2017-05-10
Added a min-width to the sub navigation and added a full height util

sroy3:
The min-width on sub navigation is so it does not change between tabs unless too large.

The full height mod class is needed because sometimes flex containers are being cut at the end of the window and do not grow to include all of the content.

### DIFF
--- a/scss/components/sub-navigation.scss
+++ b/scss/components/sub-navigation.scss
@@ -2,7 +2,9 @@
   overflow: auto;
 
   background-color: $white;
-  border-right: $sub-navigation-border;
+  border-right: $sub-navigation-min-width;
+
+  min-width: 200px;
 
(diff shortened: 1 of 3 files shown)


scss/components/sub-navigation.scss
```diff
@@ -2,7 +2,9 @@
   overflow: auto;
 
   background-color: $white;
-  border-right: $sub-navigation-border;
+  border-right: $sub-navigation-min-width;
+
+  min-width: 200px;
```
pochretien: You need to use the $sub-navigation-min-width here :)

sroy3: Haha, thanks for noticing. Replaced.

PR #355, opened 2017-05-17
Feature ADUI-2559 collapsible section in caps

dcarmichael1:
Collapsible components are now in Caps, as stated in the styleguide.


dcarmichael1:
fixed it

sroy3:
Will this affect tooltips that are in the collapsible header?

pochretien:
Tested and the tooltips stay in normal case
image

PR #359, opened 2017-05-25
*fixes input value not being displayed in Safari

sbelzile-Coveo:


scss/controls/input-field.scss
```diff
@@ -38,6 +38,8 @@
 
       border-bottom-color: $medium-grey;
 
+      -webkit-text-fill-color: $medium-grey;
```
GermainBergeron: Since we use autoprefixer you probably shouldn't use the `-webkit-` prefix

https://github.com/coveo/vapor/blob/master/gulpTasks/sass.js#L28

sbelzile-Coveo: Donc, tu écrirais directement `text-fill-color`?

GermainBergeron: Well, nevermind me. It's not supported by autoprefixer and it won't be https://github.com/postcss/autoprefixer/issues/822

PR #362, opened 2017-06-02
add a class to center the splash box if we use a splash title

pochretien:


scss/variables.scss
```diff
@@ -252,6 +252,7 @@ $splash-medium-height: 45vh;
 $splash-wizard-width: 1025px;
 $splash-wizard-padding-top: 145px;
 $splash-large-wizard-width: 1400px;
+$splash-top-bottom-padding: 70px;
```
sroy3: I would rename this `$splash-vertical-padding` since it isn't necessarily used for both but either or.

PR #370, opened 2017-06-26
+ Add style for danger and small blankslate

rousselm4:
For the strict validation setting panel


scss/components/blankslate.scss
```diff
@@ -17,4 +17,26 @@
 
     color: $medium-blue;
   }
+
+  &.mod-small {
+    padding: $blankslate-padding-small;
+    h1 {
+      margin-top: 0px;
```
GermainBergeron: Put the variables in variables.scss

rousselm4:
Applied Germain's comments

PR #372, opened 2017-07-20
Make each side of the split layout independent

sroy3:
Demo: https://nimbus.everhelper.me/client/notes/share/1023645/woaelq755ogseyxu6akf


MichelML:
nice!

PR #394, opened 2017-09-28
Added a variant to place the generic form error below something else instead of above

sroy3:
screen shot 2017-09-28 at 1 34 30 pm


MichelML:
nice

PR #396, opened 2017-10-16
Limit box style + sass format

sroy3:
screen shot 2017-10-16 at 9 55 50 am

Still waiting on a new color for gray labels since they're a bit hard to read.

### DIFF
--- a/docs/_components/components/19_limit_box.html
+++ b/docs/_components/components/19_limit_box.html
@@ -1,0 +1,64 @@
+---
+type: component
+title: Limit Box
+usage: Visually show the usage of a limit
+stylesheet: '/scss/components/limit-box.scss'
+---
+<form class="coveo-form">
+    <div class="limit-box mb2">
+        <div class="limit-box-main p2 pb1">
+            <label class="form-control-label">Limit</label>
+            <div class="limit-box-numbers pt1 flex">
+                <div class="limit-box-usage">
+                    <label class="form-control-label">Usage</label>
+                    <span class="limit-box-usage-value">42</span>
+                </div>
+                <div class="limit-box-limit form-group input-field validate">
+                    <input id="FirstLimit" type="number" value="100" required />
+                    <label for="FirstLimit">Limit</label>
+                </div>
+            </div>
+        </div>
+        <div class="limit-box-footer">
+            <div class="limit-box-bar progress-42"></div>
+        </div>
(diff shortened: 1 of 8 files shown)


sroy3:
Label color was updated 

screen shot 2017-10-16 at 10 47 25 am

docs/_components/components/19_limit_box.html
```diff
@@ -0,0 +1,64 @@
...
+                    <span class="limit-box-usage-value">100</span>
+                </div>
+                <div class="limit-box-limit form-group input-field validate">
+                    <input id="SecondLimit" type="number" value="110" required />
+                    <label for="SecondLimit">Limit</label>
+                </div>
+            </div>
+        </div>
+        <div class="limit-box-footer">
+            <div class="limit-box-bar progress-91"></div>
```
GermainBergeron: The progress-x is always an integer?

scss/components/limit-box.scss
```diff
@@ -0,0 +1,51 @@
...
+@for $i from 0 through 100 {
+  .limit-box-bar.progress-#{$i} {
+    width: $i * 1%;
+
+    @if $i >= 75 {
+      background-color: $orange;
+    }
+    @if $i == 100 {
+      background-color: $red;
+    }
```
GermainBergeron: I wonder if we should use `mods` instead. In some cases maybe 100 is better than 0

docs/_components/components/19_limit_box.html
sroy3: You fill it up with the rounded percentage. I should have added an example without a bar also, but it's possible to have a box without usage and without a progress bar.

scss/components/limit-box.scss
sroy3: That's the way I did it at first, but it felt like more work for the user to add the mod. With a mod you have to chose when to use them. This way, there are default values, but you can always change them the same way you would have with a mod. I could add 3 mods to easily overwrite the color.

sroy3: I added the mods and more examples.

sroy3:
Looks like my latest commit isn't shown here. Here is what I added: https://github.com/coveo/vapor/commit/3428413e618dbf71a203400060e3d0c50f59f78e
I'll make sure everything is there when merging and releasing.

PR #398, opened 2017-11-06
Add breadcrumb

pochretien:
screen shot 2017-11-06 at 5 01 23 pm


scss/utility/layout.scss
```diff
@@ -57,11 +57,19 @@
   min-height: 100vh;
 }
 
-.full-content {
+.full-content-x {
```
MichelML: tu pourrais commenter que full-content est deprecate mais rajouter full-content-x? sinon ca va break ou c'est utiliser right?

sroy3: It is not deprecated, it's a few lines below (same as it was), he only added 2 more classes.

scss/icons/icons.scss
```diff
@@ -17,6 +22,14 @@
     vertical-align: -2px;
   }
 
+  &.mod-20 {
+    width: 20px;
+    height: 20px;
+
+    line-height: 16px;
+    vertical-align: -2px;
+  }
```
sroy3: You only have 2 mods with specific sizes so it might be a little overkill, but you could loop through your possible mod values and only have the content once with width and height changing.

scss/icons/icons.scss
```diff
@@ -17,6 +22,14 @@
     vertical-align: -2px;
```
sroy3: You should place those two lines in 
`.mod-16, .mod-20 {...}`

scss/utility/layout.scss
MichelML: all good

PR #400, opened 2017-11-13
implement small action containers for tables

MichelML:
screen shot 2017-11-13 at 12 02 14 pm


scss/tables/table-actions.scss
```diff
@@ -102,6 +102,26 @@
   &.mod-align-header .coveo-table-actions:last-child {
     padding-right: $header-padding;
   }
+
```
pochretien: It's possible to do something with flat css ?

MichelML: can you be a bit more precise about what you want? you want me to remove its scoping under `coveo-table-actions-container` or you want more than that? @sroy3 feel free to add some details too

MichelML: not so sure it's better/useful here, since coveo-table-actions-container should always be utilized with it. I could rename `coveo-table-actions-container-small` to emphasize the link between the two if you want.

sroy3: It is not a problem of always being used with or not. If we later want to add some classes on `.coveo-table-actions` to change one property, we will have to write it like so:
```
.coveo-table-actions, .small-actions-container .coveo-table-actions {
...
}
```

Also it makes the compiled css much bigger than it needs to be.

I think you might even be able to move it out of `.coveo-table-actions-container` also.

MichelML: your statement is only true in the case of overlapping rules, `small-actions-container` is destined to be an extension of coveo-table-actions-container, much like would do in certain frameworks https://getmdl.io/components/index.html#layout-section (eg, mdl-layout, mdl-layout--fixed-header, and in our case that would be coveo-table-actions-container, coveo-table-actions-container--small). In other words, if you change the basic rules of `coveo-table-actions` under `coveo-table-actions-container`, it will also be applied to `small-actions-container` unless you decide to override this rule intentionally.

MichelML: we may be starting with different paradigms in mind

sroy3: Cascading css has a lot of disadvantage and we want to try to use it as less as we possibly can. You even added a section for it in our codestyle (https://github.com/coveo/typescript-cloudplatform/tree/master/css_sass#nested-selectors). By making it as flat as possible, it gets easier to maintain, harder to override by error, easier to build on, and we get a smaller css file. There are also a lot of reading material on the BEM's homepage that can list a lot of other advantages (http://getbem.com/introduction/).

If it is not possible to have your selectors completely flat, there are at least a few things you can do to get closer to it. You could at the very least remove the `&` from `&.small-actions-container` and move it out, but I bet you could go further than that.

MichelML: did the suggested reading, removed some nesting. note that i need to scope under `small-actions-container` because in react-vapor i don't have access to classes of every child elements.

PR #401, opened 2017-11-17
add the mod-small-content

pochretien:
add the mod-small-content to fix the height if the right content is too small


pochretien:
Before:
image

After: 
image

PR #402, opened 2017-11-21
Added limit history button

sroy3:
screen shot 2017-11-21 at 11 43 37 am


MichelML:
nice!

scss/components/limit-box.scss
```diff
@@ -62,3 +62,10 @@
 .limit-box-bar.mod-red {
   background-color: $red;
 }
+
+.limit-history-button {
+  position: relative;
+  top: -2px;
```
GermainBergeron: instead of a relative position can we use align-items on the flex-container?

sroy3: I tried, but it is not quite perfect because of line-height.

PR #403, opened 2017-11-22
! Fix width on dropdown menu items

vseguin:
This width was causing troubles in V2, so i removed it. It seems to work well, but feel free to raise a warning if you think this might break something.


vseguin:
@sroy3 It was causing issues notably if you check the dropdown in the Settings section for the product name.

PR #404, opened 2017-11-22
+ Add border-radius reset for Chrome on Mac

vseguin:


MichelML:
next time you can add Germain and Mik to the reviewers and post a screen shot of what it does before/after to help us understand, although it is simple I know.

MichelML:
I think you should use the $border-radius variable if you can @vseguin , this is the radius utilized in the btn.scss file. I might be wrong though others could confirm.

GermainBergeron:
@MichelML In this case we shouldn't use the variable because thats not how it was by default and we don't want to cause more issues

MichelML:
alright @GermainBergeron

PR #405, opened 2017-11-22
remove border of table actions feature

MichelML:
screen shot 2017-11-22 at 5 55 20 pm
  
useful to do manual coupling of actions that should be linked together when they are withing different `coveo-table-actions` scope and you have no control on that situation.

### DIFF
--- a/scss/tables/table-actions.scss
+++ b/scss/tables/table-actions.scss
@@ -94,6 +94,18 @@
   border-left: $table-border;
 }
 
+.coveo-table-actions {
+  &.mod-no-border-left,
+  &.mod-no-border-right {
+   border-left: none;
+  }
+}
+
+.coveo-table-actions.mod-no-border-right {
+  border-right: none;
+}
+
+
 .coveo-table-actions.predicate-filters + .coveo-table-actions:last-child {
   padding-left: 0;
 
--- a/scss/tables/table-actions.scss
+++ b/scss/tables/table-actions.scss
@@ -101,11 +101,6 @@
   }
 }
 
-.coveo-table-actions.mod-no-border-right {
-  border-right: none;
-}
-
-
 .coveo-table-actions.predicate-filters + .coveo-table-actions:last-child {
   padding-left: 0;
 
--- a/scss/tables/table-actions.scss
+++ b/scss/tables/table-actions.scss
@@ -95,9 +95,12 @@
 }
 
 .coveo-table-actions {
-  &.mod-no-border-left,
+  &.mod-no-border-left {
+    border-left: none;
+  }
+
   &.mod-no-border-right {
-   border-left: none;
+    border-right: none;
   }
 }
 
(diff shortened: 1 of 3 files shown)


scss/tables/table-actions.scss
```diff
@@ -94,6 +94,13 @@
   border-left: $table-border;
 }
 
+.coveo-table-actions {
+  &.mod-no-border-left,
+  &.mod-no-border-right {
```
GermainBergeron: `mod-no-border-right` should remove the border-right, no?

scss/tables/table-actions.scss
```diff
@@ -94,6 +94,13 @@
...
+   border-left: none;
```
vseguin: Weird -left:none on no-border-right?

scss/tables/table-actions.scss
```diff
@@ -94,6 +94,13 @@
   border-left: $table-border;
```
sroy3: If you change the selector here from `.coveo-table-actions + .coveo-table-actions` to `.coveo-table-actions:not(:first-child)` you might be able apply the desired border mods without having to add new rules for them.

scss/tables/table-actions.scss
MichelML: sorry will change

scss/tables/table-actions.scss
MichelML: the goal is to do manual style arrangement, not to implement a fixed rule for everything, i feel that what you're proposing will have side effects on the currently implemented tables.

sroy3: No, this will simply make the border rule from the `.coveo-table-actions + .coveo-table-actions` easier to override. You're currently adding mods to your element and they're not doing anything because `.coveo-table-actions + .coveo-table-actions` has more weight than `.mod-no-border`. If you change `.coveo-table-actions + .coveo-table-actions` to  `.coveo-table-actions:not(:first-child)` it won't be a problem anymore and the selector stays the same.

Your solution is only a patch, we should solve the problem at the root instead.

MichelML: ok i get what you mean now, will try that would be nice

MichelML: screen shot 2017-11-23 at 12 07 52 pm  
  
screen shot 2017-11-23 at 12 09 10 pm

MichelML: @sroy3 both have the same weight and thus does not fix the problem, I tried to do a more generalized hack to make our utils have more weight ( `:first-child.mod-no-border-left`, `:not(:first-child).mod-no-border-left`) , they have the same weight as `.coveo-table-actions:not(:first-child)`, but 1) not sure we want to start doing this and 2) they won't work unless we load all our utils at the bottom of `guide.scss`.

sroy3: Then I would add a specific rule on the first element or simply add the mod class on it.

MichelML: is it not what i'm doing (add the mod class on it)?

PR #407, opened 2017-12-06
add loading style table

MichelML:
screen shot 2017-12-05 at 7 07 18 pm

### DIFF
--- a/docs/_components/tables/10_table-loading.html
+++ b/docs/_components/tables/10_table-loading.html
@@ -1,0 +1,84 @@
+---
+type: tables
+title: Loading table
+usage: A style to display while table data is retrieved.
+stylesheet: '/scss/tables/table.scss'
+---
+
+<table class="no-pointer">
+    <thead class="no-pointer">
+    <tr>
+        <th class="selection">
+            <label class="coveo-checkbox-label">
+                <input type="checkbox" class="coveo-checkbox">
+                <button title="Select All" type="button"/>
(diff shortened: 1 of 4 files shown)


sroy3:
As @GermainBergeron said, usually we have an overlay with the loading animation.

MichelML:
this is not what we have in the fields table and in the log browser if you have a look, the table body becomes less opaque, pretty much it.

MichelML:
same for activity browser

sroy3:
That is because the style must be broken. Before there was an overlay with the loading anim (the background might change opacity as well or the overlay as a transparent white background). If you look at the facets, that's how it was. Also the behavior is loading row on initial load and overlay + loading anim on other loads.

MichelML:
@sroy3 I had a thorough look at this, I suspect the loading anim never has been in the tables...it is on the facets but not on the tables...that being said I moved your log-browser/act-browser rules from v2 to here for the `.loading-component` , i was abstract enough, and depending on if you put a spinner inside your container or not, a spinner should now overlay the loading component

sroy3:
Why do you need to add this when there is already `mod-loading-content`?

MichelML:
mod-loading-content is not in vapor, and is useless if you have a loading-component class , and if you want a spinner, which you asked for, this is the rules you need

MichelML:
screen shot 2017-12-06 at 11 48 00 am

PR #410, opened 2017-12-13
+ Add zendesk source icon

vseguin:
image
Pretty much straight forward.


resources/icons/svg/source-zendesk.svg
```diff
@@ -0,0 +1 @@
+<svg enable-background="new 0 0 18 16.8" viewBox="0 0 18 16.8" xmlns="http://www.w3.org/2000/svg"><path d="m8.3 5.2v10.1h-8.3zm9.7-3.5h-8.3v10.1zm-18 0c0 2.3 1.9 4.2 4.2 4.2s4.2-1.9 4.2-4.2zm18 13.6c0-2.3-1.9-4.2-4.2-4.2s-4.2 1.9-4.2 4.2z" fill="#04363d"/></svg>
```
GermainBergeron: I would be better to put the path on it's own line, also is the `16.8` really necessary? Can we use 17?

GermainBergeron: https://www.zendesk.com/company/brand-assets/#brand

You may not use the Z on its own. 😱

vseguin: Very good point - i'll ask @otaillon about it

vseguin:
image
@GermainBergeron new version with proper name

PR #412, opened 2017-12-14
Added more range to whitespace util and improved option picker style

sroy3:
I always thought the whitespace util was pretty limited, so I increased its range. Instead of entering manual margin or padding values when `p2` or `m2` was not enough, we can now use values from 1 to 5 to keep the same scale and have better uniformity.


sroy3:
You are right. I've updated it as well.

PR #413, opened 2017-12-15
revert to old font-weight for navigation

MichelML:


MichelML:
screen shot 2017-12-15 at 11 41 52 am

preserved the thin style in all browsers/OS combination, while fixing discrepancies in windows based browsing

PR #425, opened 2018-01-16
Removed height: 100%

sroy3:
This makes the border full length (see SAML screen as an example of the problem).


MichelML:
the `height` in the SAML screen is not applied. Can you post a before after here?

PR #427, opened 2018-01-18
move labeled value style to its own component

MichelML:
i will need this outside of a table in ML (see first image, in progress), other images are proof it does not break anything, also added an svg with viewBox 0 0 18 18 as it fits the specs better for ML

screen shot 2018-01-18 at 2 01 53 pm

screen shot 2018-01-18 at 2 00 16 pm

screen shot 2018-01-18 at 1 47 48 pm


scss/components/labeled-value.scss
```diff
@@ -0,0 +1,42 @@
+.box {
+  &.padded {
+    padding: 5px $collapsible-box-padding;
```
rousselm4: Ya moyen d'utiliser ou créer une variable pour les valeurs de px?

rousselm4: Nevermind, je viens de remarquer c'est juste un move, tu peux laisser ça comme ça

PR #428, opened 2018-01-19
add style for step progress bar

MichelML:
for a react-vapor component that will be needed in ml

screen shot 2018-01-19 at 12 16 57 pm

screen shot 2018-01-19 at 12 41 15 pm


scss/components/step-progress-bar.scss
```diff
@@ -0,0 +1,23 @@
+.step-progress-bar-container {
+    display: flex;
+    align-items: flex-start;
+}
+
+.step-progress-bar {
+  height: 8px;
```
rousselm4: variable?

PR #434, opened 2018-01-31
add the mod-min-height-header

pochretien:
Set a default height if the text is empty
run sass:format

### DIFF
--- a/scss/typography/headings.scss
+++ b/scss/typography/headings.scss
@@ -1,6 +1,6 @@
 h1, h2, h3, h4, h5, h6 {
   font-family: $base-font-family;
-  line-height: 1.5;
+  line-height: $header-line-height;
 }
 
 h1, .h1 {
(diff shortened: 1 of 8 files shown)


scss/variables.scss
```diff
@@ -4,6 +4,7 @@ $default-border: 1px solid $medium-grey;
 $big-title-font-size: 24px;
 $default-line-height: 1;
 $banner-height: 60px;
+$header-line-height: 1.5;
```
sroy3: You cannot set the unit here instead?

scss/typography/headings.scss
```diff
@@ -1,6 +1,6 @@
 h1, h2, h3, h4, h5, h6 {
   font-family: $base-font-family;
-  line-height: 1.5;
+  line-height: $header-line-height;
```
GermainBergeron: You can probably use 1.5em here too

pochretien: confirm, it works

PR #440, opened 2018-02-08
Clear selection button

gdostie:
Added styling for a "Clear" button in the datepicker.

### DIFF
--- a/scss/components/calendar.scss
+++ b/scss/components/calendar.scss
@@ -23,7 +23,7 @@
 
 .date-picker-box {
   width: $date-picker-width;
-  
+
   background-color: $pure-white;
   border: $default-border;
   border-radius: $border-radius;
@@ -307,36 +307,36 @@
   }
 }
 
-%date-selection-button {
-  box-sizing: border-box;
-  display: block;
-  
-  background: none;
-  border: $default-border;
-  border-radius: $border-radius;
-  
-  cursor: pointer;
-}
 
-.clear-selection-button {
-  @extend %date-selection-button;
-  color: $medium-blue;
-  font-size: $button-font-size;
-  width: 100%;
-  padding: $dropdown-choices-top-bottom-margin;
+.date-selection {
+  .action-btn {
+    box-sizing: border-box;
+    display: block;
+    
+    background: none;
+    border: $default-border;
+    border-radius: $border-radius;
+    
+    cursor: pointer;
+  }
 
-  &:hover {
-    background-color: rgba($light-blue, 0.2);
+  .action-big {
+    color: $medium-blue;
+    font-size: $button-font-size;
+    width: 100%;
+    padding: $dropdown-choices-top-bottom-margin;
+    
+    &:hover {
+      background-color: rgba($light-blue, 0.2);
+    }
+  }
+
+  .action-small {
+    width: $date-picker-dimension;
+    height: $date-picker-dimension;
+    padding: $button-small-padding-x;
   }
 }
-
-.date-button {
-  @extend %date-selection-button;
-  width: $date-picker-dimension;
-  height: $date-picker-dimension;
-  padding: 10px;
-}
-
 
 .date-separator {
   font-size: $default-font-size;
--- a/scss/components/calendar.scss
+++ b/scss/components/calendar.scss
@@ -324,6 +324,10 @@
   font-size: $button-font-size;
   width: 100%;
   padding: $dropdown-choices-top-bottom-margin;
+
+  &:hover {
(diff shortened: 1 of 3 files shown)


scss/components/calendar.scss
```diff
@@ -23,7 +23,7 @@
 
 .date-picker-box {
   width: $date-picker-width;
-
+
```
MichelML: remove whitespace

scss/components/calendar.scss
```diff
@@ -307,20 +307,37 @@
   }
 }
 
-.date-button {
+%date-selection-button {
```
MichelML: je viens d'apprendre c'etait quoi https://sass-lang.com/documentation/file.SASS_REFERENCE.html#placeholders   :)  

plus serieusement je pense que tu pourrais juste utiliser cette classe la et les autres, pas besoin de placeholder:

dans un case tu fais:

`date-selection-button clear-selection-button`  
et l'autre  
`date-selection-button date-button`

gdostie: Isn't it a DRYer solution to use a placeholder here?

gdostie: That way, no extra unecessary class will be rendered to the CSS file.

scss/components/calendar.scss
```diff
@@ -307,20 +307,37 @@
...
+  &:hover {
+    background-color: rgba($light-blue, 0.2);
+  }
+}
+
+.date-button {
+  @extend %date-selection-button;
+  width: $date-picker-dimension;
+  height: $date-picker-dimension;
+  padding: 10px;
```
MichelML: regle general est toujours de faire des variable avec les pixels pour que si tu as a faire des calculs eventually tu puisses reutiliser la valeurs...tu devrais l'ajouter sous `variables.scss` under `// Calendar & Date Picker`

scss/components/calendar.scss
MichelML: It is, it was to stay consitent with the code style of the repo in general...let's see what others think...I am not totally against it

sroy3: It'd say, leave it if it weren't for the fact that we try to avoid `@extend`. We prefer adding a class to each button instead. That way it's more obvious where the style comes from and easily reusable elsewhere without having to write another `@extend %date-selection-button` for its style.

I'm not a fan of the `@extend` at all, and I simply never use it, but it would fit more in a context of styling a webpage or a simple application than in a style guide where what we want is to give the possibility of styling by adding classes.

gdostie: I'm a big fan of using placeholders and I think we should embrace them because they generate a very dry CSS output. That beeing said, I agree since this is a style guide I shouldn't be using one in this particular use case.

scss/components/calendar.scss
```diff
@@ -307,18 +307,35 @@
...
-.date-button {
-  box-sizing: border-box;
-  display: block;
-  width: $date-picker-dimension;
-  height: $date-picker-dimension;
-  padding: 10px;
 
-  background: none;
-  border: $default-border;
-  border-radius: $border-radius;
```
MichelML: *Forget this if you are using classes already present to style your components*
```scss
  .date-btn {
    box-sizing: border-box;
    display: block;
    
    background: none;
    border: $default-border;
    
    cursor: pointer;
  }

  .date-btn-big {
    color: $medium-blue;
    font-size: $button-font-size;
    width: 100%;
    padding: $dropdown-choices-top-bottom-margin;
    
    &:hover {
      background-color: rgba($light-blue, 0.2);
    }
  }

  .date-btn-small {
    width: $date-picker-dimension;
    height: $date-picker-dimension;
    padding: $button-small-padding-x;
  }

```
no need for date-selection

MichelML: but as discussed, flattening helps if we want to overwrite the style afterwards in certain contexts

MichelML: discussed with @gdostie about this, the rules are fine like this 👍

scss/components/calendar.scss
```diff
@@ -312,13 +312,36 @@
...
+  border-radius: $border-radius;
+  
+  cursor: pointer;
+
+  &:hover {
+    background-color: $date-picker-button-hover-bg;
+  }
+}
 
+.clear-selection-button {
```
gdostie: I still think we should be using a placeholder for `.date-button` and `.clear-selection-button`.

gdostie:
I still think we should be using a placeholder for `.date-button` and `.clear-selection-button`

PR #450, opened 2018-02-28
Fixed multiline input style + validation style

gdostie:
screenshot from 2018-02-28 11-49-03

### DIFF
--- a/scss/controls/input-field.scss
+++ b/scss/controls/input-field.scss
@@ -5,7 +5,7 @@
 
   // General Styles
 
-  > input, > textarea {
+  > input, > textarea, .input-wrapper > input {
     width: 100%;
     height: $input-height;
     padding: $input-padding;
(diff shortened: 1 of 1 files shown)


sroy3:
Does the numeric spinner "With an error" and checkboxes "Validation on checkbox" still look fine?

scss/controls/input-field.scss
```diff
@@ -172,6 +173,21 @@
     }
   }
 
+  // For multiline input validation feedback
+  .input-wrapper.validate > input {
+    & + label {
+      &:after {
+        margin-top: $input-message-position * 1.5;
```
pochretien: why this 1.5 multiplicator ?

gdostie:
@sroy3 est-ce que c'était comme ça avant ?

image

image

scss/controls/input-field.scss
gdostie: Without it:

multiline

With it:

multiline2

PR #451, opened 2018-02-28
Add count (number value) to facets

ThibodeauJF:
screen shot 2018-02-28 at 4 44 11 pm

Facets are more a little bit akin to search ui, facet name that are too long get an ellipsis at the end and the count is now available. Had to refactor the style a bit to make them all fit on the same line.

### DIFF
--- a/scss/components/facet.scss
+++ b/scss/components/facet.scss
@@ -4,6 +4,7 @@
   max-width: $facet-column-max-width;
   padding: $facet-column-padding-y $facet-column-padding-right $facet-column-padding-y $facet-column-padding-left;
   margin-right: 0;
+
   border-right: 1px solid $medium-grey;
 }
 
@@ -28,10 +29,10 @@
 
   .facet-header {
     position: relative;
-  
+
     padding: $button-padding-x;
     clear: both;
-  
+
     background: $white;
     border-top-right-radius: $border-radius;
     border-top-left-radius: $border-radius;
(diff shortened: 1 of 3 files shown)


scss/components/facet.scss
```diff
@@ -17,173 +16,187 @@
...
+    clear: both;
+  
+    background: $white;
+    border-top-right-radius: $border-radius;
+    border-top-left-radius: $border-radius;
+
+    .facet-header-eraser {
+      position: absolute;
+      top: 0;
+      right: 25px;
```
MichelML: i know the code base region you're in was not a good example for what i'm about to say, but we have a rule in vapor which is to have all `px` and `%` values in variables, this way if you need to make calculations based on these values, you can reuse them. Not needed for `translateY(-50%);` but for everything else it should be done

scss/components/facet.scss
```diff
@@ -17,173 +16,187 @@
...
+    max-height: $facet-more-height;
+    padding: $dropdown-choices-top-bottom-margin 0;
+    margin: 3px 0;
+    overflow-x: hidden;
+    overflow-y: auto;
+  
+    list-style: none;
+  
+    background-color: $pure-white;
+    border: 1px solid $medium-grey;
```
MichelML: you can use `mod-border` IIRC

MichelML:
@ThibodeauJF my previous comments have been outdated with recent updates, but they might still apply

scss/components/facet.scss
```diff
@@ -17,173 +17,189 @@
...
   border-radius: $border-radius;
-}
 
-.facet-header {
-  position: relative;
+  &:last-child {
+    &.facet-opened {
+      .facet-search-results {
+        top: auto;
+        bottom: 60px;
```
sroy3: This should be a variable

sroy3: All pixel values in this file should all be variables.

ThibodeauJF:
Values now in variables

PR #458, opened 2018-03-07
Feature badges

gdostie:
Added a new component: `badge`

image


scss/components/badge.scss
```diff
@@ -0,0 +1,14 @@
+.badge {
+  display: inline-block;
+  box-sizing: border-box;
+
+  background-color: $orange;
+  color: black;
```
GermainBergeron: color should probably be $pure-white by default

gdostie: It wasn't that visible with the default orange background color, but I'm open to other suggestions for the background color

gdostie: I agree, updated screenshot thanks

PR #465, opened 2018-03-16
Feature flippable component

gdostie:
Implemented a generic flippable component

flippable2

image


resources/js/scripts.js
```diff
@@ -145,11 +145,16 @@ $(document).ready(function() {
...
         var flatSelectEl = optionEl.parent();
 
         flatSelectEl.find('.flat-select-option:not(.selectable)').addClass('selectable');
         optionEl.removeClass('selectable');
     });
+
+    // Handle flippable flip/unflip
+    $('.flippable').click(function () {
+        $(this).find('.flipper').toggleClass('show-back').toggleClass('show-front');
+    });
```
GermainBergeron: Does this needs to be duplicated in the JsAdmin too?

PR #472, opened 2018-03-21
add search-bar style

MichelML:
search-bar


MichelML:
@gdostie I guess it isn't normal that your flippable.scss was deleted?

scss/typography/utility.scss
```diff
@@ -46,8 +46,8 @@
 }
 
 .break-word {
-  word-break: break-all; // Fallback for IE11
   word-wrap: break-word;
+  word-break: break-all; // Fallback for IE11
```
sroy3: I think the order here is important

MichelML: this is not the same rule, you're sure? csscomb does this automatically and there is no straightforward way of ignoring certain lines as of now https://github.com/csscomb/csscomb.js/issues/328

otaillon:
Le text semble légèrement pas aligné au centre ? Verticalement je parle

MichelML:
@otaillon could it depend on the character entered? is it still of to you for instance:  
  
screen shot 2018-03-21 at 1 51 36 pm

otaillon:
@MichelML it is still off
screen shot 2018-03-21 at 13 54 08

scss/typography/utility.scss
GermainBergeron: I thought it was the same property but I think it's alright with the change Chrome put the word on the next line if it fit and breaks it otherwise. IE breaks it if there's no available space on the current line

scss/components/search-bar.scss
```diff
@@ -0,0 +1,65 @@
...
+  &:focus {
+    outline: 0;
+  }
+}
+
+.search-bar-icon-container {
+  float: right;
+  margin: $search-bar-icon-margin;
+
+  &, & svg {
```
GermainBergeron: you can remove the `&` in `& svg`

scss/components/search-bar.scss
```diff
@@ -0,0 +1,67 @@
...
+    width: $search-bar-icon-size;
+    height: $search-bar-icon-size;
+  }
+
+  & svg {
```
GermainBergeron: same thing here

scss/components/search-bar.scss
```diff
@@ -0,0 +1,67 @@
...
+}
+
+.search-bar-disabled svg {
+  fill: $light-grey;
+}
+
+.search-bar-loading .search-bar-icon-container {
+  position: relative;
+
+  transform: translate(-2px, -2px) scale(1) translate(2px, 2px);
```
GermainBergeron: This is a bit weird, AFAIK it does nothing?

MichelML: it does indeed nothing, removed and applied your other comments, thanks

MichelML:
@otaillon and me discussed the alignment thing and it has been solved

PR #474, opened 2018-03-23
Fix svgs for react vapor

gdostie:


gulpTasks/svg.js
```diff
@@ -21,7 +21,8 @@ function Dictionary(from) {
         const that = this;
         _.each(_.keys(this.json), (key) => {
             const camelizedKey = s.camelize(key);
-            code += '        ' + camelizedKey + ': { name : \'' + camelizedKey + '\', render : function(svgClass, spanClass, title, attr) { return svgWrapper(' + JSON.stringify(that.json[key]) + ', svgClass, spanClass, title, attr); } }, \n';
+            const svgString = JSON.stringify(that.json[key]);
+            code += '        ' + camelizedKey + ': { name : \'' + camelizedKey + '\', svgString : ' + svgString + ', render : function(svgClass, spanClass, title, attr) { return svgWrapper(' + svgString + ', svgClass, spanClass, title, attr); } }, \n';
```
GermainBergeron: instead of outputting the svg twice (once in svgString and the other in the render fct), which will nearly double the file size, could we modify the render function to use `svgWrapper(VaporSVG.svg[' + camelizedKey + '].svgString', ... ` ?

gdostie: yes, good idea

PR #478, opened 2018-03-26
Fix flippable z-index transition

gdostie:
When flipping the flippable in react-vapor, the z-index of the flippable itself is being bumped higher so that the backside is showed on top. However the transition between those 2 z-index was not smooth. This PR fixes this transition.


MichelML:
double TIL

PR #488, opened 2018-04-16
Change the style of home-cards

sroy3:
screen shot 2018-04-16 at 10 37 17 am

### DIFF
--- a/docs/_components/components/17_home-card.html
+++ b/docs/_components/components/17_home-card.html
@@ -4,12 +4,12 @@ title: Home Card
 stylesheet: '/scss/components/home-card.scss'
 ---
 
-<div style="width: 500px;">
-    <div class="home-card">
+<div class="home-cards">
+    <div class="home-card home-card-half-size">
         <div class="home-card-header">
             <h2>Home card with padding</h2>
         </div>
-        <div class="home-card-body p2">
+        <div class="home-card-body home-card-padded">
             <p>
                 Lorem ipsum dolor sit amet, consectetur adipiscing elit.
                 Aenean varius ante orci, malesuada laoreet tortor pharetra vel.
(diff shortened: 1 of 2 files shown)


docs/_components/components/17_home-card.html
```diff
@@ -4,12 +4,12 @@
...
 
-<div style="width: 500px;">
-    <div class="home-card">
+<div class="home-cards">
+    <div class="home-card home-card-half-size">
         <div class="home-card-header">
             <h2>Home card with padding</h2>
         </div>
-        <div class="home-card-body p2">
+        <div class="home-card-body home-card-padded">
```
gdostie: why not using the padding utility classes anymore ?

scss/components/home-card.scss
```diff
@@ -1,20 +1,81 @@
...
+  margin-right: 0;
+  margin-left: 0;
+
+  .home-card-header {
+    text-align: center;
   }
 }
+
+.home-card-half-size {
+  width: calc(50% - 0.5em);
```
gdostie: What does `0.5em` represent? Is it half of the padding ?

docs/_components/components/17_home-card.html
sroy3: Because you'd have to add it every time you want to use the home card and that could result in discrepancies from one card to the other. The card should either be padded or not. The padding values should not change.

scss/components/home-card.scss
sroy3: Margins between cards is `1em`. This is divided between both cards.

docs/_components/components/17_home-card.html
gdostie: Isn't it the same problem adding the `home-card-padded` every time instead of `p2` ? `p2` always has the same padding values aswell and is shorter to write.

scss/components/home-card.scss
gdostie: Perhaps a variable for the margin between cards would be more suitable then, you could reuse it in this calculation.

scss/components/home-card.scss
```diff
@@ -1,20 +1,81 @@
...
+  justify-content: space-between;
+}
+
 .home-card {
   margin-bottom: $home-card-margin;
-
+  border-radius: $border-radius;
   background: $pure-white;
   border: $default-border;
   border-radius: $border-radius;
```
GermainBergeron: border-radius is there twice

PR #489, opened 2018-04-16
Modal inside modal same width

sroy3:
React vapor modal can be inside another modal and the modal shouldn't just open inside the other modal, but open as if it were outside.

### DIFF
--- a/scss/components/modal.scss
+++ b/scss/components/modal.scss
@@ -19,9 +19,27 @@ $icon-size: (4em / 3);
   bottom: 0;
   left: 0;
 
-  width: 100%;
-  height: 100%;
+  width: 100vw;
+  height: 100vh;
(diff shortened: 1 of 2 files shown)


scss/components/modal.scss
```diff
@@ -19,9 +19,27 @@ $icon-size: (4em / 3);
...
 
-  width: 100%;
-  height: 100%;
+  width: 100vw;
+  height: 100vh;
   overflow: hidden;
+
+  .modal-backdrop, .modal-container {
+    .modal-content {
+      left: -27.5%;
```
GermainBergeron: Maybe you should create a variable for the offset on the left (2.5%) and use `left: -25% - $nested-modal-offset-left`. Same for the other sizes

PR #490, opened 2018-04-16
Fix step progress bar width

francoislg:
The progress bar in the implementation does not have any width, but in React-Vapor, the `full-content-x` class is applied, which adds a `width: 100%` property that is not rendered well at all:

Each step takes 100% while they should take equal parts from the container, thus the `flex: 1`.

So it also requires a change in React-Vapor, which is coming soon. Here is the result:

broken stepprogressbar

fixed stepprogressbar

### DIFF
--- a/scss/components/step-progress-bar.scss
+++ b/scss/components/step-progress-bar.scss
@@ -10,6 +10,8 @@
   margin: 0;
 
   border: 0;
+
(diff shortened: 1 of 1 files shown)


MichelML:
makes sens, way better implementation than full-content-x

PR #491, opened 2018-04-16
Dropdown search append and disabled option

MichelML:
append style ("APP")  
screen shot 2018-04-16 at 3 52 53 pm
  
disabled style
screen shot 2018-04-16 at 3 47 00 pm

### DIFF
--- a/scss/controls/dropdown.scss
+++ b/scss/controls/dropdown.scss
@@ -25,11 +25,29 @@
   color: $medium-grey;
 }
 
+.dropdown-option-append {
+  display: inline-block;
+  width: $dropdown-option-append-width;
+}
+
+.dropdown-selected-value {
+  display: inline-block;
+  width: 100%;
+  overflow: hidden;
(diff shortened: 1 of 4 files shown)


scss/controls/dropdown.scss
```diff
@@ -77,6 +104,20 @@
   }
 }
 
+%dropdown-option-line-style {
+  display: inline-block;
+  padding: 8px $button-padding-x;
```
sroy3: This should be a variable

scss/controls/dropdown.scss
```diff
@@ -77,6 +104,20 @@
...
+  overflow: hidden;
+
+  color: $dark-grey;
+  font-size: $dropdown-choice-font-size;
+  line-height: 16px;
```
sroy3: I think we have a variable for that.

MichelML: good will look it up

PR #498, opened 2018-04-20
Fixed width for facet column

sroy3:
Although it isn't set, the facet column minimum width is always 280px. It only fluctuates as 30% between 280px and 360px (or 80px). It is also not a constant 30% because it is calculate with a flex layout. It would be easier to calculate the fixed actions width if it weren't for that 80px dynamic sizing.


sroy3:
Problem is caused by table width and not facet column width

PR #504, opened 2018-04-27
Added Multi Step Bar component

francoislg:
Component that is more verbose that the simple "StepProgressBar".

Used to show the Status of each step of a multi-step process.

I have most certainly done some bad things here, don't hesitate to throw all your comments at me :)

### DIFF
--- a/scss/components/multi-step-bar.scss
+++ b/scss/components/multi-step-bar.scss
@@ -4,11 +4,14 @@
     .multi-step-bar {
       min-height: $multi-step-small-height;
     }
+    .multi-step-bar-text {
+      padding: $multi-step-bar-mod-small-text-padding;
+    }
   }
   &.mod-multi-step-bar-in-progress-sliding-animation {
     .multi-step-bar-doing {
       background-image: repeating-linear-gradient(90deg, transparent, transparent 11px, $multi-step-animation-color 10px, $multi-step-animation-color 25px);
-      background-size: $multi-step-sliding-animation-background-size 50px;
+      background-size: $multi-step-sliding-animation-background-width $multi-step-sliding-animation-background-height;
       background-repeat: repeat-x;
       animation: slide-background .4s linear infinite;
       z-index: 1;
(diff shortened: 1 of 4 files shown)


MichelML:
@francoislg can you post a screenshot so we can have a visual idea of your feature, we have no live demo for vapor PRs

francoislg:
image

scss/components/multi-step-bar.scss
```diff
@@ -0,0 +1,137 @@
...
+}
+
+.multi-step-bar-backdrop-container {
+  color: transparent;
+  pointer-events: none;
+  overflow: hidden;
+  .multi-step-bar {
+    transform: skewX(-20deg);
+    &:first-child {
+      padding-left: 10px;
```
MichelML: you can use the variable $default-margin everywhere here

scss/variables.scss
```diff
@@ -504,3 +504,15 @@ $search-bar-spinner-border: 2px solid $medium-blue;
 $search-bar-spinner-border-radius: 50%;
 $search-bar-spinner-offset: 6px;
 $search-bar-spinner-size: 16px;
+
+// Multi Step Bar
+$multi-step-small-height: 2px;
+$multi-step-bar-text-padding: 8px 5px;
+$multi-step-bar-separation: 5px;
+$multi-step-sliding-animation-background-size: 28px;
```
MichelML: you should name $multi-step-sliding-animation-background-size to $multi-step-sliding-animation-background-width and have another one $multi-step-sliding-animation-background-height of 50px too which you would use here https://github.com/coveo/vapor/pull/504/files#diff-e9e5407f26ebf88f698f422f2cd409a8R11

scss/components/multi-step-bar.scss
```diff
@@ -0,0 +1,137 @@
...
+  }
+}
+
+.multi-step-bar {
+  opacity: 1;
+  flex: 1;
+  display: flex;
+  align-content: center;
+  align-items: center;
+  min-height: 10px;
```
MichelML: we use variables for everything using pixel units, you should make a variable for that

PR #506, opened 2018-05-03
Side nav improvements per UX request

MichelML:
what changed:  
  
font-size is now 15px everywhere    
height of the sub-section was adjusted to UX team's taste  
there is a 1px space to the right of the scrollbar to distinguish it from the main section  

the rest was adjusted for the loading and alignment to fit with the changes above
  
I also added a -20px on the navigation tag to give more estate for the long sub-navigation name, in the admin's case "Incoherent events" would have otherwise appeared on two lines.  
   
screen shot 2018-05-03 at 1 19 29 pm

screen shot 2018-05-03 at 1 25 57 pm

### DIFF
--- a/scss/components/home-card.scss
+++ b/scss/components/home-card.scss
@@ -7,14 +7,15 @@
 
 .home-cards-padding {
   position: fixed;
-  height: 100vh;
+
   width: 100%;
+  height: 100vh;
 }
 
 .home-card {
-  display: flex;
   flex-flow: column;
 
+  display: flex;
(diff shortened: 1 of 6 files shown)


GermainBergeron:
Should the NEW tag and the collapse icons be aligned?

align-icons

MichelML:
@GermainBergeron it was not the case before but I can ask UX

MichelML:
@GermainBergeron good idea and confirmed with UX, implemented it

PR #512, opened 2018-05-09
Added an hover indicator for the arrow

sroy3:
http://nimb.ws/P9HytX
http://nimb.ws/gxwbqi


sroy3:
@MichelML I'll try what you are suggesting. It simply wasn't on what the UX showed me.

MichelML:
@sroy3 worth confirming with the UX if we really want this though!

PR #513, opened 2018-05-14
Correctly stack modals between react and backbone

sroy3:


sroy3:
@gdostie here is a screencast http://nimb.ws/yWG82N

PR #522, opened 2018-05-31
refactor side nav scroll bar

MichelML:
kapture 2018-05-31 at 14 03 12

### DIFF
--- a/docs/_sass/layout/_styleguide.scss
+++ b/docs/_sass/layout/_styleguide.scss
@@ -1,3 +1,7 @@
+.navigation, .page-content {
+  max-height: 100vh;
+}
+
 .sg-component {
   margin: 2em 0;
   border: 1px solid #ececec;
--- a/scss/components/navigation.scss
+++ b/scss/components/navigation.scss
@@ -9,12 +9,6 @@
 
   @include slim-scroll($purple-blue, rgba(255,255,255,0.5));
   &::-webkit-scrollbar, ::-webkit-scrollbar {
-    display: none;
-  }
-  &:hover::-webkit-scrollbar, :hover::-webkit-scrollbar {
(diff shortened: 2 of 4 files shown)


otaillon:
Et si tu sélectionnes une section, est-ce que le background de l'élément sélectionné se rend jusqu'au contenu ?

docs/_sass/layout/_styleguide.scss
```diff
@@ -1,3 +1,7 @@
+.navigation, .page-content {
+  max-height: 100vh;
+}
```
MichelML: this is for the demos in firefox ⬆️ otherwise the scroll does not appear, and this is not due to my changes it's already like that in http://coveo.github.io/vapor/#

MichelML: we do not have this problem in the admin I've double checked

scss/components/navigation.scss
```diff
@@ -1,12 +1,27 @@
...
-  overflow: auto;
+  overflow-x: auto;
+  overflow-y: overlay;
 
   background-color: $navigation-background-color;
 
   user-select: none;
 
-  @include slim-scroll($purple-blue, $light-grey);
+  @include slim-scroll($purple-blue, rgba(255,255,255,0.5));
```
MichelML: inspiration for the below code http://jsfiddle.net/smutnyleszek/t4q7o36r/

docs/_sass/layout/_styleguide.scss
otaillon: screen shot 2018-05-31 at 14 11 36
I seem to have it

MichelML:
discussed all the issues with @otaillon and were adressed

PR #526, opened 2018-06-07
allow touch scroll for ipads

MichelML:


scss/components/modal.scss
```diff
@@ -221,6 +221,7 @@ $icon-size: (4em / 3);
   flex-grow: 1;
 
   overflow: auto;
+  -webkit-overflow-scrolling: touch;
```
MichelML: this allows scroll on some touch devices  
  
https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-overflow-scrolling

MichelML: mostly safari on ios, which is the bug we are currently fixing

PR #528, opened 2018-06-19
Changed the action icons' size

sroy3:


MichelML:
you should probably change those under `small-actions-container` too, more precisely the value of table-small-svg-width

PR #529, opened 2018-06-21
Make thead sticky

sroy3:
screen shot 2018-06-21 at 10 15 00 am

### DIFF
--- a/scss/tables/table.scss
+++ b/scss/tables/table.scss
@@ -46,8 +46,12 @@ table:not(.datepicker-table) {
     border-top: $table-border;
     border-bottom: $table-border;
 
+    position: sticky;
+    top: 0;
+    background: $pure-white;
+    z-index: 2;
+
     &.admin-sort {
-      position: relative;
 
       cursor: pointer;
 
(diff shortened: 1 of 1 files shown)


scss/tables/table.scss
```diff
@@ -46,8 +46,12 @@ table:not(.datepicker-table) {
     border-top: $table-border;
     border-bottom: $table-border;
 
+    position: sticky;
+    top: 0;
+    background: $pure-white;
```
gdostie: `background-color` ?

sroy3: I think you're right. If there were a background-image, I wouldn't want to overwrite it. I'll change it.

scss/tables/table.scss
```diff
@@ -46,8 +46,12 @@ table:not(.datepicker-table) {
     border-top: $table-border;
     border-bottom: $table-border;
 
+    position: sticky;
```
GermainBergeron: position sticky is not supported on IE, https://developer.mozilla.org/en-US/docs/Web/CSS/position#Browser_compatibility do we have a fallback?

sroy3: I think this is just a big plus for users, but if it isn't there, it should not change their experience. I'll make sure everything still looks good in IE, but I'd opt for leaving as is if it looks the same as before and doesn't stick. We have this property elsewhere as well. I'd rather have something as trivial as this not work in one browser than jumping through hoops to make it work and risk bugs for other browsers.

GermainBergeron: I'm 100% with you on this

PR #537, opened 2018-07-03
Fix color-dots aligment

gdostie:
- Fixed flashing color dots animation (was translating upwards while scaling making it unaligned)
- Added examples for color-dots

image

### DIFF
--- a/docs/_components/utility/04_color-dot.html
+++ b/docs/_components/utility/04_color-dot.html
@@ -5,23 +5,23 @@ usage: Display a status.
 stylesheet: '/scss/elements/color-dot.scss'
 ---
 <p class="p1">Standard color dots</p>
-<i class="color-dot" style="height: 16px; width: 16px;"></i>
-<i class="color-dot state-critical" style="height: 16px; width: 16px;"></i>
-<i class="color-dot state-major" style="height: 16px; width: 16px;"></i>
-<i class="color-dot state-info" style="height: 16px; width: 16px;"></i>
-<i class="color-dot state-disabled" style="height: 16px; width: 16px;"></i>
-<i class="color-dot state-waiting" style="height: 16px; width: 16px;"></i>
+<i class="color-dot"></i>
+<i class="color-dot state-critical"></i>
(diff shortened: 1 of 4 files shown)


scss/animations.scss
```diff
@@ -29,10 +29,10 @@
 
 @keyframes scaleoutCentered {
   0% {
-    transform: scale(0.0) translateY(-50%);
+    transform: scale(0.0);
   }
   100% {
-    transform: scale(1.0) translateY(-50%);
+    transform: scale(1.1);
```
sroy3: Why did you add `.1`?

gdostie: Since opacity is 0 at 100%, it appeared smaller than the non-flashing color dots. By adding .1 it seems the same size.

docs/_components/utility/04_color-dot.html
```diff
@@ -0,0 +1,27 @@
...
+<p class="p1">Standard color dots</p>
+<i class="color-dot" style="height: 16px; width: 16px;"></i>
+<i class="color-dot state-critical" style="height: 16px; width: 16px;"></i>
+<i class="color-dot state-major" style="height: 16px; width: 16px;"></i>
+<i class="color-dot state-info" style="height: 16px; width: 16px;"></i>
+<i class="color-dot state-disabled" style="height: 16px; width: 16px;"></i>
+<i class="color-dot state-waiting" style="height: 16px; width: 16px;"></i>
+
+<p class="p1">Flashing color dots</p>
+<i class="color-dot state-executing" style="height: 16px; width: 16px;"></i>
```
vseguin: Since you always add those, should we add the height and width in the CSS directly?

scss/animations.scss
vseguin: I think it is worth a small comment as it is quite random at a glance.

docs/_components/utility/04_color-dot.html
gdostie: I agree it would make sense, I'll release a minor and make sure this rule doesn't override any existing style in our other repos.

PR #538, opened 2018-07-04
Add a style for exclude facets

acollin1:
This is a new style fore new feature on facets in react-vapor. To exclude values in facets

### DIFF
--- a/scss/components/facet.scss
+++ b/scss/components/facet.scss
@@ -88,7 +88,6 @@
 
   display: flex;
   padding: $facet-value-spacing 0;
-  overflow: hidden;
 
   font-size: $default-font-size;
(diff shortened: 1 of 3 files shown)


scss/components/facet.scss
```diff
@@ -126,6 +125,86 @@
...
 }
 
+.text-exclude {
+  text-decoration: line-through;
+}
+
+.facet-exclude {
+  padding-right: $facet-exclude-padding-right;
+
+  .exclude-icon {
```
sroy3: We try to avoid nesting style as much as possible. With such a representative name for the class, it shouldn't need to be nested inside `facet-exclude`. It might be not possible though, since our old style was abusively nested. In that case, nevermind.

PR #542, opened 2018-07-11
Remove hover on collapsible rows

gdostie:
Collapsible rows shouldn't have a hover effect by default, because when they contain big contents it doesn't make any sense to hover over everything.

### DIFF
--- a/scss/tables/table.scss
+++ b/scss/tables/table.scss
@@ -119,7 +119,7 @@
       padding-left: 10px;
     }
 
-    &:not(.no-hover):hover td {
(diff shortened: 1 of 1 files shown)


scss/tables/table.scss
```diff
@@ -119,7 +119,7 @@
       padding-left: 10px;
     }
 
-    &:not(.no-hover):hover td {
+    &:not(.no-hover):not(.collapsible-row):hover td {
```
GermainBergeron: Can you put the class `no-hover` on the row instead? This rule is getting more and more complex

PR #544, opened 2018-07-12
Fix z index table header

pochretien:


gdostie:
Make sure that datepickers and such still display properly when uses as table predicates in the admin, because last time we played with z-index it broke.

PR #558, opened 2018-07-25
Add most common list style types

gdostie:
Added multiple list style types, we had only the standrad bullets one.

image

image

### DIFF
--- a/scss/typography/lists.scss
+++ b/scss/typography/lists.scss
@@ -15,46 +15,16 @@
   }
 }
 
-%list {
-  padding-left: $list-padding;
+$list-style-types: disc, circle, square, decimal, lower-alpha, upper-alpha;
 
-  li {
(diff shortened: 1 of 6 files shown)


scss/typography/lists.scss
```diff
@@ -0,0 +1,60 @@
...
+}
+
+%list {
+  padding-left: $list-padding;
+
+  li {
+    margin-bottom: $box-spacing;
+  }
+}
+
```
GermainBergeron:
```sass
$list-styles: decimal, circle, disc, square, lower-alpha, upper-alpha;

@each $list-style in $list-styles {
  .list-#{$list-style} {
    list-style: $list-style;
    @extend %list;
  }
}
```
In fact you could remove the %list and put the style directly in the loop

gdostie: nice, was looking for this thx

PR #570, opened 2018-08-14
Rounded borders utilities

gdostie:
Added utilities for rounded borders, also added a demo page for borders utilities.

image

### DIFF
--- a/scss/utility/border.scss
+++ b/scss/utility/border.scss
@@ -31,3 +31,27 @@ $borders: border-top border-bottom border-left border-right;
 .mod-no-border {
   border: none;
 }
+
+@for $i from 1 through 10 {
+  $radius: $border-radius * $i;
+
+  .mod-rounded-border-#{$i} {
+    border-radius: $radius;
+  }
+  
+  .mod-rounded-border-top-#{$i} {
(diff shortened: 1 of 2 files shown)


scss/utility/border.scss
```diff
@@ -31,3 +31,27 @@ $borders: border-top border-bottom border-left border-right;
 .mod-no-border {
   border: none;
 }
+
+@for $i from 1 through 10 {
```
MichelML: add units https://stackoverflow.com/a/11275156

gdostie: Where do I add the units ? `$i` is just a multiplicator of the `$border-radius` variable which already has `px` units.

sroy3:
We usually use a border radius of 2. Is there a use for the other ones?

gdostie:
@sroy3 yes, new UX mockups sometimes include a radius of `4px`, so instead of just adding this one I added some more just in case.

image

scss/utility/border.scss
pochretien: dont need, $border-radius already have px and your $i is only a multiplicator

MichelML: sorry my mistake nvm @gdostie

PR #574, opened 2018-08-15
Added a max-width to the color-dot so it does not get squished on flex layouts

sroy3:
Also ran sass format


sroy3:
I'm not sure if it could ever get squished vertically, but I'll add it. Better safe than sorry.

PR #575, opened 2018-08-15
Ensuring Vapor keeps control of placeholder color in IE11

samisayegh:
Screenshot of the issue we have inside the Interface Editor when using IE11.

visible_placeholder_ie11

### DIFF
--- a/scss/lib/_mixins.scss
+++ b/scss/lib/_mixins.scss
@@ -20,12 +20,27 @@
 }
 
 @mixin placeholder($text-color: $medium-grey, $font-family: $base-font-family) {
-  &::placeholder {
-    color: $text-color;
+  &::placeholder, &:-ms-input-placeholder {
     font-family: $font-family;
     font-size: inherit;
     text-transform: none;
-
+  }
+  
+  &::placeholder {
+    color: $text-color;
     transition: color 0.2s ease;
+  }
+
+  &:-ms-input-placeholder {
+    // IE11 treats input placeholders like input text. This bug has been known for some time but has yet
+    // to be fixed.
+    //
+    // The !important is necessary to ensure the Vapor CSS styles regarding the placeholder are applied.
+    // If the keyword is removed, and a Vapor user customizes the colour of the input text,
+    // the placeholder will always be visible in IE11 and will overlap with a label if present.
+    //
+    // Adding !important allows consumers of Vapor to customize the colour of input text without affecting
+    // the visibility of the placeholder inside IE11.
+    color: $text-color !important;
   }
 }
--- a/scss/lib/_mixins.scss
+++ b/scss/lib/_mixins.scss
@@ -20,27 +20,23 @@
 }
 
 @mixin placeholder($text-color: $medium-grey, $font-family: $base-font-family) {
-  &::placeholder, &:-ms-input-placeholder {
+  &::placeholder {
+    color: $text-color;
     font-family: $font-family;
     font-size: inherit;
     text-transform: none;
-  }
-  
-  &::placeholder {
-    color: $text-color;
+    
     transition: color 0.2s ease;
   }
 
   &:-ms-input-placeholder {
-    // IE11 treats input placeholders like input text. This bug has been known for some time but has yet
-    // to be fixed.
-    //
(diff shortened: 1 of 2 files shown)


sroy3:
We use autoprefixer to fix this problem on our side. The generated CSS correctly adds the `:-ms-input-placeholder` and we have another autoprefixer in our project when importing SCSS files directly. We try not to add any vendor prefixes directly in the stylesheets. Is this something that could work for you, or is there some other problem I am not aware of?

scss/lib/_mixins.scss
```diff
@@ -20,12 +20,27 @@
...
     text-transform: none;
-
+  }
+  
+  &::placeholder {
+    color: $text-color;
     transition: color 0.2s ease;
   }
+
+  &:-ms-input-placeholder {
```
MichelML: if you have a url to a stackoverflow discussion or else explaining this situation, you could provide it in this comment.

scss/lib/_mixins.scss
```diff
@@ -20,12 +20,27 @@
 }
 
 @mixin placeholder($text-color: $medium-grey, $font-family: $base-font-family) {
-  &::placeholder {
-    color: $text-color;
+  &::placeholder, &:-ms-input-placeholder {
```
MichelML: following what @sroy3 explained here https://github.com/coveo/vapor/pull/575#issuecomment-413300881 , you don't need to add &:-ms-input-placeholder since autoprefixer does it for you. Apart from that, I do think all of your changes will fix the issue at hand.

scss/lib/_mixins.scss
```diff
@@ -20,12 +20,27 @@
...
     font-family: $font-family;
     font-size: inherit;
     text-transform: none;
-
+  }
+  
+  &::placeholder {
+    color: $text-color;
     transition: color 0.2s ease;
```
MichelML: will it cause problem if the transition rule is applied to `:-ms-input-placeholder`  ?  

if so, add `transition-property: none;` under :-ms-input-placeholder at line 34

MichelML:
@sroy3 I do think the change is overall valuable here because there are rules that must be different depending on the pseudo-class  (::placeholder vs :-ms-input-placeholder)

scss/lib/_mixins.scss
samisayegh: Unfortunately, there isn't one thread that captures the situation fully:

Regarding my point of IE11 treating input placeholders like input text, it comes from empirical observation that the `color` attribute affects the placeholder in colour in IE but not other browsers.

As relates to the fix, these two threads suggest using the !important keyword:
- https://stackoverflow.com/questions/22199047/placeholder-css-not-being-applied-in-ie-11
- https://stackoverflow.com/questions/40118053/placeholder-not-working-with-internet-explorer

I think I will include the first link since it is more descriptive.

scss/lib/_mixins.scss
samisayegh: Ok, good point, in that case I will remove the second selector, and move the ::placeholder `color` and `transition` properties back into this block.

scss/lib/_mixins.scss
samisayegh: I separated these two properties out initially because when I added `&:-ms-input-placeholder` on line 23, I was no longer able to override the color in a stand-alone `&:-ms-input-placeholder` block on line 34. I moved the `transition` into this block too because moving just the `color` property broke the transition.

I will move both back into the main block on line 23.

samisayegh:
@sroy3 Hi Stephanie, following your feedback, I removed the vendor prefix attached to `::placeholder`. 

However, I kept the stand-alone `:-ms-input-placeholder` because as @MichelML pointed out, there is a difference in the `color` property between IE11 and other browsers. In the case of IE11, we need to include the `!important` keyword for the reasonsI left in my comment.

scss/variables.scss
```diff
@@ -26,10 +26,8 @@ $title-font-size: 16px;
 $medium-title-font-size: 18px;
 $big-font-size: 24px;
 $big-title-font-size: 32px;
-
```
MichelML: this is me who just ran our formatter

MichelML:
thanks for the fix @samisayegh  👍

PR #580, opened 2018-08-28
Added variables for the status cards

sroy3:


scss/variables.scss
```diff
@@ -580,11 +580,18 @@ $diff-separator-margin: 10px;
...
 $full-banner-height: 300px;
 $banner-content-height: 165px;
 $banner-content-padding: 50px;
 $banner-right-width: 300px;
 $banner-warning-icon-fix: 2px;
 $banner-box-shadow: inset 0 5px 10px 0 rgba($black, 0.1), inset 0 -10px 50px 0 rgba($black, 0.2);
+
+// Status Cards
+$status-card-border-width: 4px;
+$status-card-padding: 25px;
```
GermainBergeron: is this supposed to be padding-x?

sroy3: It's used more generally than just horizontal padding. In a way `status-card-padding-y` could have been ` $status-card-padding / 2`, but I didn't want half pixels.

PR #581, opened 2018-08-29
truly working responsive side nav

MichelML:
this version works flawlessly in the admin, demo incoming  

sidenav2


scss/components/header.scss
```diff
@@ -6,20 +6,41 @@
...
   }
 }
 
+
 .header-height {
   height: $panel-header-height;
 }
+
+.header-hamburger {
+  padding: 15px 20px;
```
sroy3: Don't forget to move this to variables

scss/components/header.scss
```diff
@@ -6,20 +6,41 @@
...
+
+  background-color: transparent;
+  border: none;
+
+  outline-width: 0;
+  cursor: pointer;
+
+  transition: opacity $navigation-toggle-duration ease-in-out;
+
+  opacity: 0.7;
```
sroy3: I think there are variables for opacity. It's `$transparency-1(...)` or something like that.

GermainBergeron: There is some variable for transparencies, see variables.scss

scss/components/header.scss
GermainBergeron: I hate to be this guy, but it should be variables

scss/variables.scss
```diff
@@ -226,6 +226,14 @@ $navigation-loading-height: 12px;
 $navigation-loading-item-border-radius: 10px;
 $navigation-loading-background-size: 300% 104px;
 $navigation-loading-bullet-width: 12px;
+$navigation-toggle-duration: 0.3s;
+%navigation-toggle-closed {
+  margin-left: -$navigation-width;
+  transition: transform $navigation-toggle-duration ease-in-out; 
+}
+%navigation-toggle-opened {
```
GermainBergeron: Those (%navigation-toggle-closed and %navigation-toggle-opened) are used in v2?

MichelML: yes for status and quick nav which are outside the nav bar

PR #587, opened 2018-09-11
Revert backdrop size for prompts

sroy3:
A prompt backdrop is inside another view and often this other view is a modal, so the size should be the one of its parent and not of the whole page.


MichelML:
@sroy3 do you have an example of what you want to do? not sure of how it would look like?

sroy3:
For the example, see the ADUI-3641 in Jira. This fixes the bug (how it was before).

PR #589, opened 2018-09-12
add coveo child vertical

MichelML:
this is for the small arrow you see on the container at your right

screen shot 2018-09-12 at 3 44 05 pm


MichelML:
this is for the small arrow you see on the container at your right

PR #600, opened 2018-10-16
Increase select dropdowns z-index value

gdostie:
Had this issue with a SingleSelectConnected in the actions bar of a table (which is sticky with a z-index of 1)

image

Do you think this could cause any issue elsewhere?

### DIFF
--- a/scss/controls/dropdown.scss
+++ b/scss/controls/dropdown.scss
@@ -238,7 +238,7 @@
 }
 
 .select-dropdown-container {
-  z-index: 2;
+  z-index: $table-header-z-index;
 
   box-shadow: $list-box-box-shadow;
   border: $default-border;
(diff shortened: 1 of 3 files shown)


pochretien:
Im not sure its the good way to increase the dropdown z-index. The question is, why the header of the table need to have a z-index greater than 1

GermainBergeron:
I think it would be better to set the z-index of the action bar to be the same as the table header, this way the dropdown would appear above

gdostie:
@GermainBergeron I've just checked and `th` have a z-index of 1 and `.coveo-table-actions-container` have a z-index of 2 and the dropdown still appears underneath the `th` because it has a z-index of 1 and is probably getting rendered later.

gdostie:
The only thing that works for now is the fix that I made here.

GermainBergeron:
I think it would be best to add a variable for the sticky-header z-index and use it here

PR #604, opened 2018-11-06
put point of coveo child vertical middle left of container

MichelML:
was previously at the top of the container

screen shot 2018-11-06 at 5 10 57 pm

only important file https://github.com/coveo/vapor/pull/604/files#diff-3080f51217d354123cfb41cf57c6898d

### DIFF
--- a/scss/forms/form-child.scss
+++ b/scss/forms/form-child.scss
@@ -36,8 +36,10 @@ $child-arrow-height: 10px;
   &.vertical {
     margin-left: $child-arrow-height + $child-parent-spacing;
 
+    border-top-left-radius: $border-radius;
+
     &:before {
-      top: 0;
+      top: calc(50% - #{$child-arrow-height});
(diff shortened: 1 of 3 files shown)


otaillon:
Do we still have both ?

scss/forms/form-child.scss
```diff
@@ -36,8 +36,10 @@ $child-arrow-height: 10px;
   &.vertical {
     margin-left: $child-arrow-height + $child-parent-spacing;
 
+    border-top-left-radius: $border-radius;
+
     &:before {
-      top: 0;
+      top: calc(50% - #{$child-arrow-height});
```
GermainBergeron: It would be nice to have this style with a new class to avoid breaking places where the arrow at the top was what we wanted

MichelML: it wasn't used anywhere, i implemented it this way a couple weeks ago but figured that in the middle would look nicer

PR #619, opened 2018-12-11
fix: apply top positioning on sticky table header only if supported by browser

gdostie:
top positioning for sticky table headers will now be applied only if the browser supports sticky positioning.

https://developer.mozilla.org/en-US/docs/Web/CSS/@supports

### DIFF
--- a/scss/tables/table.scss
+++ b/scss/tables/table.scss
@@ -203,8 +203,10 @@
   text-align: right;
 }
 
-.coveo-table-actions-container + table:not(.datepicker-table) th {
-  top: $table-actions-container-height;
+@supports ((position: sticky) or (postion: -webkit-sticky)) {
+  .coveo-table-actions-container + table:not(.datepicker-table) th {
+    top: $table-actions-container-height;
+  }
 }
 
 .table-container {


scss/tables/table.scss
```diff
@@ -203,8 +203,10 @@
   text-align: right;
 }
 
-.coveo-table-actions-container + table:not(.datepicker-table) th {
-  top: $table-actions-container-height;
+@supports ((position: sticky) or (postion: -webkit-sticky)) {
```
samisayegh: My concern with this approach is that we may still have a problem in IE11? If the `top` rule is needed for the thread component, it will not be applied after this change. Here is Stephanie's initial PR: https://github.com/coveo/vapor/commit/4575a06cfee3e4b00c64ef23bc435b51b3b6d86a

gdostie: Since `sticky` is not supported by IE, it will be rendered as `static`. In static positioning, the top rule is not applied so the `top` rule was not applied in IE before this fix anyway.
> When position is set to static, the top property has no effect. (https://developer.mozilla.org/en-US/docs/Web/CSS/top)

gdostie: After reflecting about this, the top rule may not be the problem afterall. As you explained ealier, when `sticky` is not supported by the browser, it is rendered as `static` by default. However the `top` rule has no effect on statically positioned elements. Not sure what is the problem then, I'll have to test this in browserstack as I'm not on windows.

gdostie: I've just tested this and by removing the rule
```css
.coveo-styleguide .coveo-table-actions-container + table:not(.datepicker-table) th {
    top: 52px;
}
```
in IE 11, it renders properly. In the computed styles, I can see that `position: relative` is applied on the `th` rather than `static` so the `top` rules takes effect here.

samisayegh: Just to make sure, was your test with the thread component component that Stephanie was working on?

PR #624, opened 2019-01-21
Make split-layout wrap on small screens

gdostie:
Split-layouts will now wrap its columns when screen is smaller than our small screens breakpoint (992px)
Demo:
image
image


MichelML:
just make sure it doesn't break something in the admin, fyi model testing is using a splitlayout although it seems like it's not

gdostie:
@MichelML The 2 models in the model testing will stack on top of each other, but there are components that break the UI more than that in the panel IMO. So I'll merge here and we'll fix all the responsiveness problems of the model testing panel later on.

PR #651, opened 2019-03-28
New utils

pochretien:
add transparency utils for each variable
add utils for the hover with each color in the palette
add specific classes for the card


pochretien:
Examples incoming

scss/utility/layout.scss
```diff
@@ -155,3 +155,11 @@
 .small-desktop-breakpoint-responsive-UI {
   min-width: $breakpoint-desktop-small;
 }
+
+@for $i from 1 through length($transparencies) {
+  .mod-transparency-#{$i} {
```
GermainBergeron: I'm not sure layout is the best place to put this but I don't really mind

I would prefer to call it `transparency-X` (without mod-)

scss/variables.scss
```diff
@@ -528,6 +528,10 @@ $ribbon-font-size: 0.8em;
 $ribbon-letter-spacing: 1px;
 $ribbon-rotation-angle: 45deg;
 
+// Card
+$card-container-box-shadow: 0 2px 8px 0 rgba(0, 0, 0, 0.3);
```
GermainBergeron: `$card-container-box-shadow: 0 2px 8px 0 rgba($black, $transparency-4);` (I'm not sure which number it should be)

pochretien:
image
image
image

PR #658, opened 2019-04-09
Fix table header borders in Edge

francoislg:
Fixes an issue only found in Edge related to borders in tables.

Found out it was cause by `background-color: #ffffff`, and when searching on the internet, [found this stackoverflow question here](https://stackoverflow.com/questions/7517127/borders-not-shown-in-firefox-with-border-collapse-on-table-position-relative-o/16337203#16337203).

Before:

2019-04-09 14_41_46-Microsoft Edge

After:

2019-04-09 14_41_32-Microsoft Edge

Only happening in Edge, somehow.

### DIFF
--- a/scss/tables/table.scss
+++ b/scss/tables/table.scss
@@ -10,6 +10,7 @@
   th, td {
     position: relative;
 
+    background-clip: padding-box;
     text-align: left;
     vertical-align: middle;
 


francoislg:
@GermainBergeron @gdostie ? 😄

PR #663, opened 2019-04-25
Added the favicon file and the link to it

otaillon:
I can't get it to work on my local demo (apparently local favicons are hard), but it is supposed to work 😅
64x64


docs/_includes/head.html
```diff
@@ -15,6 +15,8 @@
 
     <link rel="canonical" href="{{ page.url | replace:'index.html','' | prepend: site.baseurl | prepend: site.url }}">
 
+    <link rel="shortcut icon" type="image/png" href="../../resources/images/favicon.ico">
```
GermainBergeron: I think it should be: 

```suggestion
    <link rel="shortcut icon" type="image/png" href="/dist/images/favicon.ico">
```

otaillon: What is _dist_ ? I put my favicon into _resources/images/favicon.ico_ and it appeared to have been created in _dist_ ?

GermainBergeron: _dist_ is the folder that we _distribute_, when we compile/modify/minify/... files they are moved there to avoid modifying the original file

gdostie: `dist` is a shortname for `distribution`, it is typically the name of the folder in which we put all files we want to distribute to the client.

otaillon: Should I then also move my favicon.ico to there ? Or it's always automatically copied ?

gdostie: IIRC images folder gets injected in the dist folder at compile time, so it should be good.

PR #664, opened 2019-04-26
Fixed the popover border

otaillon:
For https://coveord.atlassian.net/browse/UX-1396
Also updated the favicon link and design


scss/components/popover.scss
```diff
@@ -3,7 +3,7 @@
 
   background-color: $pure-white;
   box-shadow: 0 2px 4px rgba($stratos, 0.4);
-  border: 1px solid $dark-blue;
+  border: 1px solid $medium-grey;
```
gdostie:
```suggestion
  border: $default-border;
```

otaillon: TIL

PR #674, opened 2019-05-23
Fix adui 4421 update readme

gdostie:


gdostie:
Demo: https://github.com/coveo/vapor/blob/fix-ADUI-4421-update-readme/README.md

PR #677, opened 2019-05-29
Plug-in prettier and lint everything

gdostie:
- Removed tsjs
- Removed csscomb
- Setup prettier
- Lint source files
- Added the "code style" badge to the readme
image


.csscomb.json
```diff
@@ -1,309 +0,0 @@
-{
```
GermainBergeron: One neat thing with csscomb was the re-ordering of the properties in scss, should we keep it?

gulpfile.js
```diff
@@ -142,7 +143,16 @@ function Dictionary(from) {
...
+            code +=
+                '        ' +
+                camelizedKey +
+                ": { name : '" +
+                camelizedKey +
+                "', svgString : " +
+                svgString +
+                ", render : function(svgClass, spanClass, title, attr) { return svgWrapper(VaporSVG.svg['" +
+                camelizedKey +
+                "'].svgString, svgClass, spanClass, title, attr); } }, \n";
```
GermainBergeron: this is pretty ugly/hard to read, we should replace with a template literal

```typescript
code += `
    ${camelizedKey}: {
        name: ${camelizedKey},
        svgString: ${svgString},
        render: function(svgClass, spanClass, title, attr) {
            return svgWrapper(VaporSVG.svg['${camelizedKey}'].svgString, svgClass, spanClass, title, attr);
        } 
    },`;
```
I didn't test this ^

package.json
```diff
@@ -1,79 +1,93 @@
...
+    },
+    "husky": {
+        "hooks": {
+            "pre-commit": "lint-staged"
+        }
+    },
+    "lint-staged": {
+        "**/*.{js,scss,json,md,yml}": [
+            "npm run lintfix",
+            "git add"
```
GermainBergeron: might need `git add --staged` instead as we discussed yesterday

gdostie: `git add --patch` is what is suggested in the [prettier documentation](https://prettier.io/docs/en/precommit.html#option-1-lint-staged-https-githubcom-okonet-lint-staged).

I tried it and my git process stalled into a deadlock, leaving a file called `index.lock` somewhere in the hidden `.git` folder... I had to manually delete the lock file in order to run any git command.

`git add` seems to be working properly, I would leave it be until something breaks.

gulpfile.js
gdostie: I agree it is ugly! I think it is out of scope of this PR though, I think only lint changes should be done in this PR.

.csscomb.json
gdostie: @GermainBergeron I think [stylelint](https://github.com/stylelint/stylelint) would be a better option because prettier already [has plugins for it](https://prettier.io/docs/en/integrating-with-]linters.html#stylelint). I think it would be a cool addition to add in another PR.

